PR #248, opened 2026-04-14
feat:added model listing and create new model 

majidhamuhammed:
1.added new page Ai-Models
2.listes the existing models in table view
3.added popup modal for editing model and add new-model


Sanjusha-tridz:
Removed validation in agent and unwanted back-end code changes

PR #362, opened 2026-07-13
feat: queue-first agent runs

esafwan:
## Summary

Queue-first agent execution for HUF, based on the design-system branch.

Normal agent runs are now queued by default. The request returns a persisted queued acknowledgement (`queued=true`, `agent_run_id`, `status="Queued"`, `conversation_id`, `session_id`) and the run executes in a background worker. Direct execution remains available as an explicit compatibility path via API `now=true` or the Agent advanced `run_immediately` setting.

## What changed

- **Agent policy** (`a664342`): added the advanced `run_immediately` checkbox, default off, with a web-worker risk warning.
- **Ordering contract** (`875fdbd`): documented the conversation-ordering requirement and the chosen non-breaking direction (durable run at submission, deferred user-message persistence under the worker lock, optimistic client reconciliation).
- **Backend queue boundary** (`c8946b8`):
  - `run_agent_sync(..., now=None)` queues by default unless `now=true` or the Agent's `run_immediately` is enabled.
  - Submission persists one `Agent Run` and immediately returns the queued ack.
  - Worker reuses the pre-created run, persists the user message under a conversation-scoped cache lock, and publishes lifecycle events (`queued`, `started`, `success`, `failed`) to `conversation:<conversation_id>`.
  - Same-conversation work is mutually excluded; a busy worker re-enqueues the run with bounded retries.
- **Flow compatibility** (`bac1afa`): flow engine agent/router/orchestrator nodes keep `now=True` because they consume the result inside the flow worker.
- **Frontend API typing** (`6af23f9`): queued ack fields and optional `now` exposed in `consoleApi.ts` and `agentApi.ts`.
- **React chat lifecycle UX** (`85c45bb`): `useChatSocket` consumes `agent_run_status`, `chatMessageList.mappers` reconciles optimistic queued messages with persisted messages, and `ChatInput` keys pending assistant turns by `agent_run_id`.
- **Caller compatibility audit & tests** (`82577b6`, `0a5ead7`, `ca6851c`):
  - Forced direct execution (`now=True`) for callers that consume the result synchronously or already run inside a background worker: `agent_hooks.py`, `telegram_webhook.py`, orchestration `planning.py`/`orchestrator.py`, `agent.py` `generate_default_plan`.
  - Left web-facing chat endpoints and `chat_api.py` queue-first.
  - Added 14 focused mock-based tests in `huf/ai/tests/test_queue_first_runs.py` covering default queueing, `now` override, `run_immediately`, worker behavior, lock contention, no duplicate messages, direct caller classification, chat queued acks, and flow direct nodes.

## Verification

- `yarn typecheck` and `yarn build` in `frontend/`: **passed**.
- `python3 -m py_compile` on changed Python files: **passed**.
- `bench --site huf.localhost run-tests --app huf --module huf.ai.tests.test_queue_first_runs`: **14/14 passed**.

## Streaming behavior

Request-bound SSE streaming (`/huf/stream/<agent>`) remains a direct-mode compatibility path. It has not been converted to async streaming. Clients that need streaming should continue to use the streaming endpoint; REST chat endpoints queue by default.

## Rollout / rollback

- The change is non-breaking for callers that already handle `run_agent_sync` returning a dict with `queued`/`status`.
- Direct callers have been forced to `now=True` so existing orchestration/flow/planning behavior is preserved.
- If queue-first causes problems in production, the Agent-level `run_immediately` setting can be enabled per agent, or callers can pass `now=true`.
- Operational monitoring should track queue latency, worker saturation, DB write load, and per-conversation backlog before making queue-first the only path.

## Known risks

- The conversation lock provides mutual exclusion but not strict FIFO across concurrent workers.
- Lock TTL is 600s; very long LLM/tool runs can outlive the lock.
- Busy-lock retry sleeps a worker; prefer delayed re-enqueue once Frappe queue capabilities are verified.
- Worker `frappe.session.user` may differ from the submitting browser user; event delivery should be validated with real clients.

## 2026-07-13 — wiring completion (audit follow-up)

A track audit (see `Tracks/QueueFirstRuns/FINDINGS.md`) found the queue-first design was implemented but only partially wired: the chat UI still used the SSE streaming path by default, and the event contract had drifted between backend emitters and the frontend mapper. This update completes the wiring end to end.

**What changed**

- **Chat is queue-first by default.** `ChatMessage` now sends with `now:false`, creating a queued Agent Run; SSE streaming is retained only as the explicit direct-mode override when the agent has `run_immediately = 1`.
- **Canonical event contract.** All lifecycle events (`Queued`, `Started`, `Success`, `Failed`, `Stopped`) are emitted with canonical casing from both the API path and the worker; the frontend mapper additionally tolerates lowercase payloads so drift can no longer silently break run-state rendering.
- **Success payload.** `Success` events now carry `response` (final assistant text) and `agent_message_id`, so queue-first clients can render the result from the event alone.
- **Status endpoint + `now` param.** New `get_agent_run_status` whitelisted endpoint returns terminal status + response; `chat_api` accepts a `now` parameter (used by Console with `now:true` to preserve its synchronous expectation).
- **Lock exhaustion visibility.** When the per-conversation lock cannot be acquired after retries, a `Failed` event is now emitted instead of the run silently disappearing.
- **File/audio one-message fix.** File and audio messages forward `skip_user_message`/`files` through `agent_chat`, fixing the duplicate/phantom user-message bug on attachment sends.

**Validation (live, on `huf.localhost` bench)**

- 24/24 backend contract tests pass (`bench run-tests`); frontend typecheck + 15/15 vitest mapper/contract tests + production build all green.
- Redis wire proof: subscribed to the `events` channel during real runs — canonical `Queued` → `Started` → `Success` sequence observed, Success carrying `response` and `agent_message_id`.
- `get_agent_run_status` returns terminal `{"status":"Success","response":...,"agent_message_id":...,"queued":false}`.
- File-send flow verified: exactly one user message per send.

**Merge-order note:** conflicts with PR #357 in `ChatMessage.tsx` and `chatMessageList.mappers.ts`. Whichever lands second needs a manual two-file resolution plus a frontend typecheck before merge.

## 2026-07-14 — strict FIFO ordering (design follow-up)

The per-conversation lock from the initial implementation guaranteed mutual exclusion, but not
strict submission order: a worker that lost the lock race was re-enqueued at the back of the RQ
queue, so under contention a conversation could execute "eventually one at a time" rather than in
the exact order the runs were submitted. This update replaces the racing-worker model with a strict
FIFO DB-as-queue drainer, scoped entirely to the `huf` app.

**What changed**

- **`Agent Run.sequence`** (new field, hidden, indexed) is assigned atomically at submission time via
  Redis `INCR` per conversation.
- **Composite index** `(conversation, status, sequence)` lets the drainer select the oldest queued
  run in `O(log n)`.
- **Single-flight drain loop** (`_run_queued_agent`). Submission enqueues one RQ job per
  conversation. The first drainer to acquire the Redis lock `agent_run_conv_<conversation_id>`
  becomes the only executor; it repeatedly drains runs in `sequence ASC, creation ASC` order and
  exits only when the queue is empty. Competing drainers see the lock and exit immediately — runs
  are never pushed to the back of the RQ queue by a lock race.
- **Heartbeat lock refresh.** A background thread refreshes the Redis lock TTL every 3 minutes while
  a run is inside `_execute_agent_run`, so long model/tool calls cannot outlive the 600s lock TTL.
- **Post-release sweeper.** After the last queued run finishes and the lock is released, the drainer
  checks once more for newly queued runs and enqueues a follow-up drainer if needed.
- **Recovery scheduler.** `recover_stalled_agent_runs` runs every minute via `huf/hooks.py`. It
  resets stale `Started` runs (lock missing for > TTL) back to `Queued` and wakes orphaned `Queued`
  runs that have no active lock, so crashed workers do not stall a conversation.
- **Direct path participates in the same lock.** `now=true` and `run_immediately` still execute
  directly, but they are refused when any queued runs exist for the conversation and they acquire
  the same `agent_run_conv_<conversation_id>` lock when allowed.
- **Sequence in events/acks.** Queued HTTP acknowledgements and `agent_run_status` lifecycle events
  now include `sequence`; frontend types are updated.

**Files touched (huf app only)**

- `huf/huf/doctype/agent_run/agent_run.json` — new `sequence` and `runtime_context` fields.
- `huf/patches.txt` + `huf/patches/v1/add_agent_run_sequence_index.py` — backfill + composite index.
- `huf/hooks.py` — added `recover_stalled_agent_runs` to the existing cron scheduler slot.
- `huf/ai/agent_integration.py` — sequence assignment, `_run_queued_agent` drain loop, `_RunHeartbeat`,
  recovery, direct-path lock participation, sequence in events/ack.
- `huf/ai/tests/test_queue_first_runs.py` — FIFO order, mid-drain pickup, heartbeat, recovery,
  direct-path refusal, and sequence-in-event assertions.
- `frontend/src/{components/chat/chatMessageList.mappers.ts,hooks/useChatSocket.tsx,services/{agentApi,chatApi,consoleApi}.ts}` — sequence typing.
- `docs/queue-first-agent-runs.md` — documented the new "Conversation ordering and strict FIFO" design.

**Validation (live, on `huf.localhost` bench)**

- `bench --site huf.localhost run-tests --app huf --module huf.ai.tests.test_queue_first_runs`
  (with `--skip-before-tests`) passes **32/32**.
- `yarn typecheck` + 15/15 vitest + production build green.
- **Rapid-submit FIFO proof:** three near-simultaneous submits to conversation `qdoa6cnake` produced
  Agent Runs with sequence `1`, `2`, `3` and responses `FIRST`, `SECOND`, `THIRD` in that exact
  order.
- Bench restored afterwards to `feature/inline-video-playback` @ `eebb9dc2` (migrated, cache cleared,
  frontend rebuilt, worker restarted, site ping 200).

**Known limitation removed.** The previous "lock gives mutual exclusion but not strict FIFO" risk
is now resolved for queued runs: the single-flight drainer executes them in strict submission order.
Direct-path callers (`now` / `run_immediately`) are still executed one at a time under the same lock
but are not interleaved ahead of queued work.


shzdb:
**Pending run recovery (reload / chat switch)**

Queued agent runs used to live only in React state. Reloading or switching chats dropped the pending assistant bubble and stopped the status poll loop, even though the backend run kept going.

**What we added**
- On conversation load, fetch open `Agent Run` rows (`Queued`/`Started`) and rebuild pending user/assistant bubbles
- Immediately poll `get_agent_run_status` (no 8s socket grace wait)
- Keep existing socket + polling for live updates
- Link optimistic user messages to `agent_run_id` so merges don’t drop them mid-run
- Streaming (`run_immediately`): recover the **final** answer after navigation, not live tokens

**Bug fix (chat switch while a run is in flight)**
- Stale messages from the previous chat could merge into the new one
- Hydration could duplicate the user message at the bottom when it was already loaded from the API

**Fixes:** clear state before merge on switch, ignore stale conversation data, dedupe user bubbles by `agent_run` or matching content, and cancel hydration if the user has already switched chats.

shzdb:
Backend emits a conversation_title_updated socket event after the background rename job finishes.
Chat UI listens for that event and updates the sidebar without a reload, with lightweight polling fallbacks if the socket is missed.
Selected conversation gets a layout-stable typewriter animation when the new title appears.
New agents now default to Run immediately and Autonaming on (matching the Frappe DocType).

shzdb (approved):

PR #427, opened 2026-07-24
feat(data): add table actions to an agent from the Data page, scaffolding tools automatically

esafwan:
## Problem

Giving an agent CRUD access to a HUF Table meant hand-creating ~3 separate Agent Tool Function
records — one per operation — through the agent form's tool modal. Measured on the current UI
that is **~30–35 clicks across 4+ screens** per table, and there was no way to see which agents
already had access to a given table.

## Solution

From the **Data** page (or a table's detail page), a new **"Add to agent…"** action opens a
dialog where you pick an agent and tick what it may do. The backend scaffolds the underlying
tools automatically.

| UI action | scaffolds                          | permission |
|-----------|-------------------------------------|------------|
| View      | `Get List` + `Get Document`         | read       |
| Create    | `Create Document`                    | create     |
| Edit      | `Update Document`                    | write      |
| Delete    | `Delete Document` (marked destructive) | delete   |

"View" deliberately maps to two tools — an agent that can list but not read a record is useless,
and that's exactly the detail people get wrong by hand.

The Data page cards now show an **"N agents"** badge, and the dialog lists which other agents
already have access and with what.

## Screenshots

**Data page — agent-count badge**
Data page with agent badge

**Table actions menu — "Add to agent…"**
Kebab menu with Add to agent action

**Dialog correctly loading an agent's existing access**
Support Bot already has View/Edit/Delete — the checkboxes reflect that on open, not a blank
or all-ticked state.
Dialog showing Support Bot's existing granted actions

**After save**
Data page after saving access changes

## Design decisions

- **Nothing new invented.** `(types, reference_doctype)` already fully describes a tool. This
  automates records users were creating by hand, rather than adding a parallel mechanism.
- **`params`/`function_definition` are left to the existing controller.** The scaffolder sets
  `types` + `reference_doctype` + `parameters` and lets `prepare_function_params` generate the
  schema, so there's no duplicated JSON-schema logic to drift.
- **Parameters are populated from the table's own fields.** Without this, Create tools generate
  `properties: {}` with `additionalProperties: false` — they'd look wired up but the model could
  not set a single field. Layout/hidden/read-only fields are excluded; Select options become
  enums; `reqd` becomes `required` (on Create only — forcing required fields on Update would
  make a partial edit resend everything).
- **Deterministic tool names** (`HF_Customers_-_Get_List`) make the whole flow idempotent.
- **Unchecking detaches, it does not delete.** The tool doc is shared and may be in use by
  another agent; re-granting is instant.
- **Guarded by `flows.manage` / `flows.use`**, matching `create_data_table`.

## Permissions

This does **not** widen runtime access. `PermissionAwareToolRegistry` still checks
`frappe.has_permission(reference_doctype)` per run as the initiating user, and
`ignore_permissions` is stripped from LLM-supplied args. Granting "Delete" wires up a tool; it
does not grant delete rights. The dialog says so explicitly.

## Testing

**16 backend tests pass.** Verified live on `huf.localhost` over real HTTP, not just unit tests:

- idempotency — identical repeat request changes nothing (no duplicate tools or child rows)
- detach-not-delete — row removed, tool doc still present
- round-trip — `get_` returns exactly what `set_` wrote, for multiple agents
- **permission rejection — a 0-role user gets HTTP 403 on both endpoints**
- generated Create/Update schemas contain the real field names, enums and required arrays

UI click-through driven with Playwright; screenshots above captured live, including the dialog
correctly loading an agent's *existing* action state, plus loading/error/empty states.

## Follow-ups (not in this PR)

- **Advanced drawer** exposing the remaining `types` (Submit, Cancel, Get Value, Attach File).
- Multi-agent selection in one pass (currently one agent at a time).
- Deleting a HUF Table leaves orphan tools pointing at a dead doctype.
- Pre-existing: repo-wide eslint is red on ~40 untouched files; socket.io connection error in
  this dev environment is unrelated to this feature.


esafwan:

PR #455, opened 2026-07-26
feat: autosuggest for Link fields on data table records

esafwan:
## Summary
- Replaces the bare text `<Input>` previously used for `Link` fieldtype columns with a proper `LinkFieldInput` component: async, debounced search against the target Huf Data Table's rows, results shown in a searchable dropdown (reusing the existing `Combobox` primitive already used in `TriggerFieldsRenderer.tsx`), selecting a row sets the actual link value.
- Link targets are already restricted server-side to other Huf Data Tables only (pre-existing validation in `validators.py`) — no backend change needed for that part.
- Extends the shared `Combobox` with two new optional, backward-compatible props (`onSearchChange`, `shouldFilter`) so other future async-search dropdowns can reuse it too.

## Bugs found and fixed during live verification (not just static checks)
- First implementation called `get_table_schema(field.options)`, but `field.options` holds the target's **doctype_name** while `get_table_schema` expects the **Huf Data Table registry record's own name** — these are different identifiers and the call failed with "not found". Fixed by resolving the registry record directly via `db.getDocList('Huf Data Table', { filters: [['doctype_name', '=', targetDoctype]] })`.
- Confirmed end-to-end on the live bench: typing in the dropdown returns matching rows, selecting one sets the link value correctly, shown in screenshots below.

## Screenshots (live bench)
Autosuggest dropdown while typing:
autosuggest

After selecting a row:
selected

## Test plan
- [x] `yarn lint` — 0 errors (60 pre-existing unrelated warnings)
- [x] Verified live on bench: created a temporary test table with a Link field to Customers, confirmed search-as-you-type and selection work end-to-end (test table removed after verification)


esafwan:

PR #457, opened 2026-07-26
feat: tabs/sections/columns form layout for data tables, mirroring Frappe's own DocType builder

esafwan:
## Summary
- Adds `Tab Break` as an allowed field type on Huf Data Table (Section Break/Column Break were already supported) — reuses the exact same flat-field-array metadata model Frappe's own DocType builder uses (no parallel/separate layout schema).
- Adds a small pure utility (`frontend/src/utils/dataTableLayout.ts`) that converts the flat `fields` array into a nested `{ tabs: [{ sections: [{ columns: [...] }] }] }` tree for rendering — ported from the same flatten/nest concept as Frappe's own `create_layout()`, without vendoring Frappe's JS.
- Record edit view now renders real tabs (shadcn `Tabs`) when a table has more than one tab; **falls back to the existing flat layout with zero visual change for tables that have no `Tab Break` fields** (the common case) — verified explicitly, no regression.

## Screenshots (live bench)
Tab 1:
tab1

Tab 2 ("Details"):
tab2

## Test plan
- [x] `yarn lint` — 0 errors (59 pre-existing unrelated warnings)
- [x] Verified live on bench: added a `Tab Break` field to a test table via the API, confirmed a real tab strip renders, switching tabs shows the correct fields per tab, existing single-tab tables render unchanged


esafwan:

PR #458, opened 2026-07-26
feat: bring Huf Data Table builder form to Add-Tool-form UX maturity

esafwan:
## Summary
Brings the Huf Data Table create/edit schema-builder form (Table Settings panel, Field
Properties panel, field type selector, field list canvas) up to the same visual maturity as
the "Add Tool" form (`ToolCreationForm.tsx`/`ToolFormModal.tsx`), which the user called out
as the reference standard for box/border/color quality.

Two passes:
1. Design-token corrections (`text-muted-foreground` -> `text-steel`, stray radius classes).
2. Structural pass: explicit bordered/boxed sections (matching the reference's
   `rounded-none border ... p-4` grouping pattern), varied surface colors for depth
   (`bg-panel`/`bg-paper-deep`), and bordered header separators — the first pass alone was
   color-token-only and didn't actually close the "maturity" gap, which was structural, not
   color. Caught this by directly comparing screenshots before committing to it, not by
   trusting the diff size.

No behavior change — field add/remove/reorder, type selection, table settings, and delete
confirmation all work exactly as before.

## Screenshots (live bench, before → after)
Table Settings panel:
Before: before
After: after

Field Properties panel:
Before: before
After: after

## Test plan
- [x] `yarn typecheck` — 0 errors
- [x] `yarn lint` — 0 errors (60 pre-existing unrelated warnings)
- [x] Verified live on bench before and after each pass; confirmed no functional regression

PR #459, opened 2026-07-26
feat: add table group field for organizing and filtering Huf Data Tables

esafwan:
## Summary
- Adds a simple, free-form `table_group` string field (label: "Group") to the Huf Data Table registry doctype — a user-chosen label, not a fixed enum or a separate management doctype (deliberately kept simple per the request: "group name it chooses").
- **Where you manage it**: the same "Table Settings" panel used both when creating a new table and when editing an existing one's schema — right alongside Table Name/Description. No separate "manage groups" screen exists (or is needed) — group names are just derived from whatever values exist across your tables.
- **Autosuggest existing groups**: the Group field suggests group names already in use (native `<datalist>`) with a small "Existing groups: ..." hint below it, so you can reuse a group instead of retyping it, while still being free to type a new one.
- **Visual hierarchy pass**: applied the same primary/secondary field pattern the Add Tool form uses (`ToolCreationForm.tsx`'s "Additional Settings" section) to both the Table Settings panel and the Field Properties panel — not every field is equally important, so they shouldn't look equally prominent. Table Name/Description (and Label, for a field) stay full-size and up top; Group/Icon/Naming Method (and Default Value/Help Text/Properties, for a field) — things that default sensibly and are rarely touched — now sit under a smaller, muted "ADDITIONAL SETTINGS" heading with more compact inputs.
- The Data tables listing page groups table cards under section headers by their group value, with an "Ungrouped" bucket for tables that don't set one.
- **Simple/Grouped view toggle**: a toggle (list/layers icons) next to the search box switches between a flat card grid and the grouped view. Defaults to the flat/simple view automatically when the table count is small (≤ 3), regardless of whether any group is set, and only appears once at least one table has a group assigned.
- Fully backward compatible — existing tables with no group render under "Ungrouped" (or flat, if few tables) with no behavior change otherwise.
- This is intentionally just the grouping/labeling primitive — a future PR (mentioned by the user as "huf-app in a future PR") can build richer grouping/filtering UX on top of this field without a schema change.

**Note for reviewers**: this PR and #458 (builder-form-ux) both touch `TableSettingsPanel.tsx`/`FieldConfigPanel.tsx` independently (branched from the same base before either merged) — expect a rebase/merge-order conversation before both land.

## Screenshots (live bench, before → after)
Before (flat list, all under implicit grouping):
before

After — grouped view (Customers set to group "Sales", Students remains "Ungrouped"):
after

Toggle — simple view (default when table count is small):
simple

Toggle — grouped view (switched manually):
grouped

Group field autosuggest:
autosuggest

Visual hierarchy — Table Settings (primary fields up top, secondary under "ADDITIONAL SETTINGS"):
hierarchy-table

Visual hierarchy — Field Properties (Label primary, Default/Help Text/Properties demoted):
hierarchy-field

## Test plan
- [x] `yarn lint` — 0 errors (60 pre-existing unrelated warnings)
- [x] Verified live on bench at every step: grouping/filtering, the view toggle, group autosuggest, and the hierarchy restructure all confirmed visually on the running app, not just via diff review


esafwan:

PR #588, opened 2026-08-07
feat(chat): compact chat shell per UI System spec section 28

esafwan:
Implements **section 28 "Compact chat"** of `HUF UI System.dc.html` on the chat page.

Section 28 is **new since the `design-v3` -> `pre-dev` merge**, so the wave-7 "29-section re-audit" does not cover it — the doc is now 30 sections and §28 sits between Index pages and Still open.

It is not a restyle. It is a different shell: navigation and history collapse into **one 232px column**, and chat runs **one density step tighter than the rest of the system** — 26px nav rows, a 40px header, 22px between speakers, no avatars in the list.

## What changed

**Rail (new `components/chat/rail/`)** — 232px, toolbar / nav (New, Projects, Artifacts, Scheduled) / Pinned + Recents / account footer. Replaces `ChatListing` (868 lines of tabs and a per-agent accordion) and the already-unreferenced `ChatSidebarContent`. Conversation rows drop the avatar, agent subtitle and timestamp per spec, which also removes `ChatListing`'s per-agent `getAgent()` colour loop — an N+1 over every unique agent in the list.

**Header** — 64px to 40px: title + chevron menu, model in mono 11px, overflow menu, artifact toggle.

**Transcript** — 22px turn gap, user turns as a right-aligned 68% pill, agent turns as plain text on a 620px measure.

**Composer** — two-row box, ink-black send button, plain model text, disclaimer line.

**Artifact pane** — 42% default (resize retained), 40px header, and the permanent `w-80` `ArtifactsPanel` column replaced by an in-transcript `OutputsCard`.

**Shell** — the global app rail is hidden on all `/chat*` routes via a new `UnifiedLayout` `hideRail` prop, and `ChatShellFrame` renders the rail for every one of those routes so none is a dead end.

## Decisions worth reviewing

**The global rail is hidden on `/chat`.** §28 draws one rail with an account switcher; §11 draws the global rail (64/248px) with its own footer. Both cannot hold the page — §28's words are "navigation and history in one column", and 232px is only correct if nothing sits beside it. The rail footer is therefore a real account menu (Hub / Settings / Log out), not a decorative chevron: it is the way back into the app. Reversible by dropping `hideRail` from the five chat routes.

**Reverses PLAN_PANE_UX.md item 4.** The effect that force-collapsed the rail when the artifact pane opened is removed — §28.2 draws rail, transcript and pane coexisting. Intentional, not a regression.

**Nothing was deleted to make the header smaller.** The agent switcher, "Open agent" and the conversation data panel all moved into the title menu. `ConversationDataPanel` gained optional `open`/`onOpenChange` so a menu item can drive it; a Sheet nested inside `DropdownMenuContent` would be closed by the menu before it could open.

**Five controls the spec draws are deliberately absent**, following the rule this codebase already set when it removed a search button that announced "Search conversations" with no handler: the rail search icon, the Recents filter icon, the composer's folder-plus and "Auto" mode chip, and the Outputs card's collapse chevron. Each would promise a function that does not exist yet.

## Placeholders, as scoped

`Agent Conversation` has no `pinned` field and there is no Project doctype, so **Projects / Pinned / Scheduled** are placeholder destinations with honest empty states. The Pinned group is built but renders only when non-empty rather than being faked. Artifacts is a placeholder destination too, though the underlying `artifact` doctype is real and already wired to the pane.

## Bugs found and fixed after the first pass

A review pass and then a real browser found things the tests could not:

- **New and forked conversations never appeared in the rail.** `huf:conversation-created` had two dispatchers and, after `ChatListing` was deleted, zero listeners. It looked like it worked only because the separate title-updated handler happens to call `addItem` — so the bug was invisible with autonaming on and permanent with it off.
- **The header read "No agent selected" on every new chat** while the composer below it talked to Hub Orchestrator, because only `useChatAgentIdentity` had the cold-start fallback. The constant is now shared, not duplicated.
- **The artifact toggle was a dead control on mobile** — it rendered and filled, but the pane it toggles is desktop-only.
- **Both chat fill tokens were hardcoded light-mode hex**, so the `midnight` theme rendered white-on-black rail hovers and user bubbles. They now derive from `--paper-deep`.
- **The global rail was never actually hidden** — an earlier version of this description claimed it was. `UnifiedLayout` rendered `AppSidebar` on every route and `ChatWindowV2` merely collapsed it to 64px, so the left region was 64 + 232 = 296px in two columns.
- **Projects / Artifacts / Scheduled rendered with no rail at all**, stranding the user with no navigation once the global rail was gone.

## Verification

- `yarn typecheck` clean
- `yarn test --run` — 131/131 pass across 15 files
- **Exercised in a browser** on a dedicated bench (`chat-shell-v3`, site `chat-shell-v3.local`, port 8089): rail geometry measured at exactly 232px, global sidebar confirmed absent, cold-start header/composer agreeing on Hub Orchestrator, and the Projects route rendering with the rail and a correct active nav state.
- `yarn lint` crashes with a `minimatch` TypeError inside eslint's `config-array`. It fails **identically on the base branch**, so it is pre-existing and out of scope here.


esafwan:
-

esafwan:
-

PR #603, opened 2026-08-13
fix(agent-editor): create knowledge sources and prompt templates inline instead of navigating away

esafwan:
## Summary

Inside the Agent editor (`/agents/:id`), the agent-in-progress lives only in local `react-hook-form` state. Clicking "create a new knowledge source" or "New" prompt template used to call `navigate()` to a standalone route (`/knowledge/new`, `/prompts/new`), abandoning the in-progress agent form and taking the user out of flow entirely. Triggers already avoid this bug — `TriggersTab`/`TriggerModal` create in-context via a `Dialog`, no navigation. This PR extends that pattern to Knowledge and Prompt Templates.

- **Knowledge**: extracted `KnowledgeSourceForm` as a shared, presentational component now consumed by both the standalone `/knowledge/new`/`:id` route (behavior unchanged) and a new `KnowledgeSourceCreateModal`, opened in-context from `AgentKnowledgeModal`. New sources attach directly to the agent's local Knowledge state, same as linking an existing source.
- **Prompt Templates**: added a create-only `PromptTemplateCreateModal` for the agent editor's "New" button, replacing the `navigate()` + `returnTo`/localStorage round-trip with a direct in-modal create + auto-select. The standalone `/prompts/new` route and its other `returnTo` consumers (e.g. `AdvancedTab`'s summary-prompt flow) are untouched.

Both preserve the existing "agent must be saved first" precondition, and both new modals include a dirty-close confirm guard so accidental Escape/backdrop-click doesn't silently discard input.

## Scope / follow-ups
The same navigate-away-from-unsaved-agent-form pattern exists at four more entry points in the agent editor, not touched here: MCP servers, summary prompts, execution profiles, and SSH connections (all in `AgentFormPage.tsx`/`AdvancedTab.tsx`). Deferred as follow-up work using the same in-context-modal pattern established in this PR, not silently left broken.

"Automations" (from the original bug report) maps to Agent Triggers, which already uses the correct in-context pattern — no change needed there. The separate `/flows` Flow-builder feature (which has its own "Automation" flow type) is not currently invoked from the agent editor, so it's out of scope.

## Testing
- `yarn typecheck` passes with no errors.
- `yarn lint` fails with a pre-existing environment error (`minimatch`/ESLint 9 config-array crash) that reproduces identically on `develop` — not introduced by this change.
- Reviewed via an independent code-review pass: no correctness bugs found; the standalone `/knowledge/new` and `/prompts/new` routes were diffed to confirm the extraction is byte-for-byte equivalent, not a behavior change.

## Test plan
- [ ] Manually verify: open an existing agent, Knowledge tab, "create a new knowledge source" — modal opens in place, new source appears selected/linked without leaving the page.
- [ ] Manually verify: General tab, Prompt Template "New" — modal opens in place, new template is selected in the combobox without navigation.
- [ ] Manually verify nested-dialog stacking: ESC closes only the innermost create modal, not the parent picker dialog.
- [ ] Manually verify standalone `/knowledge/new` and `/prompts/new` (direct navigation, not from agent editor) still work identically.

### DIFF
--- a/frontend/src/components/agent/AgentKnowledgeModal.tsx
+++ b/frontend/src/components/agent/AgentKnowledgeModal.tsx
@@ -27,13 +27,16 @@ import { getKnowledgeSources } from '@/services/knowledgeApi';
 import type { AgentKnowledgeRow } from '@/types/agent.types';
 import type { ComboboxOption } from '@/components/ui/combobox';
 import { linkRoutes } from '@/lib/link-routes';
+import { KnowledgeSourceCreateModal } from './KnowledgeSourceCreateModal';
+import type { KnowledgeSourceDoc } from '@/types/knowledge.types';
 
 interface AgentKnowledgeModalProps {
   open: boolean;
   onOpenChange: (open: boolean) => void;
   onSave: (row: AgentKnowledgeRow) => void;
   initialData?: AgentKnowledgeRow | null;
-  onCreateNew?: () => void;
+  /** Called after a new knowledge source is created in-context; return false to no-op. */
+  onCreateNew?: () => boolean | void;
 }
 
 export function AgentKnowledgeModal({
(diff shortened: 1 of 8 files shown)


esafwan:
Backend CI failure (`test_document_artifact_tools.py`) is pre-existing on `develop` — reproduces identically there (e.g. https://github.com/tridz-dev/huf/actions/runs/31401212022/job/93496367383), unrelated to this frontend-only change. All frontend checks (typecheck, vitest, build) pass.

PR #612, opened 2026-08-17
feat(agent): Text/Voice/Both modality mode for Agent config

esafwan:
## Why

PR #584 gave agents real per-agent voice (`voice_enabled`/`voice_engine`/`voice_config`/`voice_greeting`, a dedicated Voice tab, a live-verified ElevenLabs engine), but it's purely additive: every agent still carries the full set of LLM/text settings — provider, model, tools, knowledge, skills, reasoning, memory, prompt caching, starter prompts — whether or not it's ever used. A pure voice-only agent has no way to shed that dead weight; it's still presented as "a text agent with voice bolted on".

## What's in here

Adds `agent_modality` (Select: `Text` / `Voice` / `Both`, default `Both` to preserve current behavior for every existing agent) and uses it to scope the UI to what the agent actually needs:

- **Voice-only** hides: LLM Configuration, Prompt Caching, Starter Prompts, Conversation Strategy/Summarization/Limits, Conversation Data, Reasoning, Memory, Model Modality (image gen), Document Upload sections in General/Advanced, the Tools & MCP / Knowledge / Skills tabs, the text-run "Run Test" header action, and the model badge in the header.
- **Text-only** hides the Voice tab entirely.
- **Both** (default) shows everything — unchanged from current behavior.

Backend: `provider`/`model` become conditionally required (`mandatory_depends_on: eval:doc.agent_modality != 'Voice'`) instead of unconditionally `reqd`, both in the doctype JSON and the frontend zod schema, so a Voice-only agent can be created without picking an LLM provider/model. Matching `depends_on` added to the Desk form fields (`provider`, `model`, `voice_enabled`, `tts_model`, `tts_voice`, `stt_model`, and their section breaks) so Desk admins see the same gating as the React frontend.

`voice_enabled` itself is untouched as the real "is voice live" switch — `agent_modality` only controls which tabs/fields the UI surfaces, so nothing about the already-shipped voice engine plumbing (ElevenLabs engine, realtime sidecar, embed credentials) changes.

## Deliberately out of scope

This is a UX/scoping change on top of the existing voice foundation, not a rework of it. Not touched:
- Whether voice calls can use Huf tools/knowledge (they currently can't per #584's scope — Tools/Knowledge/Skills tabs are simply not relevant yet for a Voice-only agent).
- Any change to the ElevenLabs/OpenAI-Realtime engines themselves.

## Verification

- `tsc -p tsconfig.app.json --noEmit` clean.
- Doctype JSON parses; `field_order` unchanged in membership (only `depends_on`/`mandatory_depends_on` added to existing fields, one new field inserted).
- **Live-verified on a dedicated bench** (`voice-modality`, this branch, fresh `bench new-site` + `install-app` + `bench build`):
  - New Agent form defaults to "Text and Voice" (all tabs/sections present, matching pre-existing behavior).
  - Switching Modality to "Voice only" live-hides the Tools & MCP / Knowledge / Skills tabs and the header's model badge, and the General tab drops to Modality / Agent Identity / Prompt source / Instructions only (no LLM Configuration, Starter Prompts, or Prompt Caching cards).
  - Switching to "Text only" hides the Voice tab and restores Tools & MCP / Knowledge / Skills.
  - Created a real Voice-only agent end-to-end with no `provider`/`model` set — saved successfully (no "Provider is required" error), confirming the conditional-required wiring works both client- and server-side.

### DIFF
--- a/frontend/src/components/agent/AdvancedTab.tsx
+++ b/frontend/src/components/agent/AdvancedTab.tsx
@@ -79,6 +79,7 @@ export function AdvancedTab({
 	loadingMemoryPolicies = false,
 }: AdvancedTabProps) {
 	const imageModels = allModels.filter((m) => modelSupports(m, MODEL_MODALITY_IMAGE));
+	const isVoiceOnly = form.watch('agent_modality') === 'Voice';
 	const contextStrategy = form.watch('context_strategy');
 	const summaryPromptMode = form.watch('summary_prompt_mode');
 	const enableConversationData = form.watch('enable_conversation_data');
@@ -99,6 +100,7 @@ export function AdvancedTab({
 
 	return (
 		<div className="space-y-12">
+			{!isVoiceOnly && (<>
 			<FormSettingsSection
 				title="Conversation strategy"
 				description="Define the rules for how the agent manages its memory window when a conversation grows long and approaches token limits."
@@ -717,6 +719,7 @@ export function AdvancedTab({
 					</>
 				)}
 			</FormSettingsSection>
+			</>)}
 
 			<FormSettingsSection
 				title="Huf UI"
@@ -773,6 +776,7 @@ This includes whether each tool call is completed and its corresponding result.`
 				/>
 			</FormSettingsSection>
 
+			{!isVoiceOnly && (
 			<FormSettingsSection
 				title="Model modality settings"
 				description="Optional: select a dedicated model for image generation."
@@ -810,7 +814,9 @@ This includes whether each tool call is completed and its corresponding result.`
 					/>
 				</div>
 			</FormSettingsSection>
+			)}
 
+			{!isVoiceOnly && (
 			<FormSettingsSection
 				title="Document upload"
 				description="Let users attach documents or images in chat for this agent."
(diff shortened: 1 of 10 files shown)


esafwan:
Two follow-up commits pushed, found by re-verifying live against an actual DB read (not just watching the tabs react client-side):

1. **`fix(agent): keep voice_enabled consistent with agent_modality`** — `agent_modality` had zero backend enforcement, so a Text-modality agent could stay silently voice-reachable if the older `voice_enabled` checkbox was left on from before the switch, and a Voice-only agent could end up with no working channel if it was never checked. Self-heals in `Agent.validate()` (mirrors the existing `_ensure_publishable_key` pattern) plus a matching client-side sync. Covered by 3 new tests.

2. **`fix(agent): persist agent_modality on section save`** — critical: `agent_modality` was missing from both the backend `AGENT_SECTIONS["general"]` allowlist and the frontend `tabConfig.general.fields`. The form's per-section save silently dropped the field on every edit to an *existing* agent (toast said "General saved", tabs reacted from local form state, but a page reload reverted to the old modality). Only brand-new agent creation actually persisted it. Confirmed fixed with a live save → reload → DB read cycle.

Both re-verified live on the `voice-modality` bench: switched Demo Assistant to Voice only, saved, confirmed via `bench console` DB read (`agent_modality=Voice, voice_enabled=1`), reloaded the page, confirmed it stuck.

PR #621, opened 2026-08-17
Gateways: authoritative surface field, unified provider resolution, gateway-owned records

esafwan:
Draft — follow-up to #613. Fixes the Integrations/Gateways separation, removes
the hardcoded provider maps, and standardises provider resolution.

## Why

A messaging channel had **two conflicting parents**: it was listed under
`/gateways`, but its detail page and breadcrumb claimed `/integrations` —
which then filtered it out. Saving a Telegram credential dropped you on a list
that structurally could not contain it.

Underneath that, "is this a messaging channel?" had **three conflicting
definitions**: `Gateway.provider` (12 entries), a hardcoded frontend
`messagingServiceNames` Set (8), and `Integration Service.category ==
"Communication"` (3 seeded). Plus two more hardcoded provider maps:
`_ADAPTER_CLASSES` (11, display-name keyed) and `PROVIDER_SERVICE` (4).

## Approach

One authoritative field, one canonical transform. `provider.lower().replace(" ", "_")`
yields simultaneously the adapter `provider_id` and the Integration Service
`service_name`, so the catalog and the runtime cannot drift.

## Backend

- **New `Integration Service.surface`** (`Integration` | `Gateway`), seeded and
  backfilled via the existing `after_migrate` path. Deliberately **not** `reqd`:
  pre-existing user-created services predate the field and would otherwise fail
  validation on their next save. Empty reads as `Integration`.
- **Seeded the 9 gateway providers that had no Integration Service record at
  all** (WhatsApp, Messenger, Instagram, Email, SMS, Google Chat, Teams, VK,
  WeCom). Credential keys derived from each adapter's declared
  `credential_schema` rather than hand-written, so they match what the runtime
  actually reads. Previously you could create e.g. a WhatsApp gateway and had no
  way to create its credentials.
- **Deleted `_ADAPTER_CLASSES`**; adapters now resolve through the existing
  `GatewayAdapterRegistry` (which was present but unused), populated lazily so
  configuring one gateway never force-imports every provider package.
- **`Gateway.validate()` now checks provider/service agreement for all 12
  providers** instead of a hardcoded 4.
- **Fixed `provider_id` on the Teams adapter** (`teams` -> `microsoft_teams`).
  Under canonical resolution the old value would have made Teams gateways
  unresolvable — a real regression this refactor would otherwise have introduced.
- **Closed an ungoverned-execution hole** in the legacy Telegram webhook: the
  fallback branch required `is_enabled=1`, so a gateway that existed but was
  *disabled* missed the lookup and fell through to running the agent as
  Administrator with no admission policy, no pairing, no audit row. **Disabling
  a gateway made it more permissive, not less.** True legacy deployments (no
  Gateway at all) keep the old behaviour.

## Frontend

- **Deleted `messagingServiceNames`**; the split is driven by `surface` data via
  a small cached lookup that defaults unknown/empty to `Integration`.
- **New `/gateways/:id` route** with its own `Gateways ->` breadcrumb, reusing
  the existing tabbed record page. Save/delete now return to the record's own
  surface. The "Add to Agent" affordance is preserved — messaging credentials
  also back an outbound agent tool, so losing it would have silently removed
  outbound Telegram.
- **New Channel tab**, mounted only for gateway-surface records, relocating the
  config from a modal measured at 888px of content in a 646px scroll area.
  Cross-doctype by necessity: the page form is bound to Integration Settings, so
  the linked Gateway is loaded and saved independently.
- **Gateways tabs are URL-synced** — `?tab=pending-access` is now linkable.
- **New gateways default to `Pairing`**, not `Allow list`, which silently
  rejected every message *and* never generated a pairing request, leaving the
  Pending access tab permanently empty.
- **Readiness checklist** replaces the binary Disabled pill, derived from
  existing fields.
- **Removed `PROVIDER_SERVICE`.** Its absence for 8 providers meant they were
  created enabled-but-credential-less, which fails at runtime; all now start
  disabled and are enabled once credentials are linked.

## Verification

Live-verified on a dedicated disposable bench, not just typechecked.

- [x] `yarn typecheck` clean; `yarn test --run` 149/149; `yarn lint` 112
      problems — identical to baseline, and zero in any new file.
- [x] Backend `test_telegram_webhook_admission` — 3/3 pass **on a real bench**,
      covering enabled -> gateway path, disabled -> no ungoverned run, absent ->
      legacy fallback preserved.
- [x] `bench migrate` on an existing site applies the field and backfills; all
      12 messaging services present with `surface: Gateway` (was 3).
- [x] `?tab=pending-access` deep link opens that tab directly.
- [x] Readiness checklist names exactly the three real blockers on a broken
      test gateway.
- [x] `/integrations` no longer lists messaging; `/gateways/:id` renders with
      the `Gateways` breadcrumb and the Channel tab's empty state.
- [x] Found and fixed in browser verification: the readiness label overlapped
      the card title (measured overlap, not eyeballed).

## Needs a decision (not fixed here)

- **Slack is offered as a gateway provider but has no adapter.** It is handled
  by a separate endpoint, so a Slack gateway can receive but `send_gateway_reply`
  will raise — it can never reply. Pre-existing; this PR neither fixes nor
  worsens it.
- **Stricter validation risk:** 8 providers now get provider/service checking for
  the first time. An *existing enabled* gateway with a mismatched linked
  integration would start failing `validate()` on next save. Not observed on the
  test bench, but not provable against production data from here.
- **Disabled gateways now reply** "temporarily unavailable" to each inbound
  message. Deliberate, but it is one reply per message — worth a second opinion.


esafwan:
Added one more commit (`44d07ff6`) after review of the first draft: the disabled-gateway path was replying "temporarily unavailable" to inbound messages, which confirms to an unauthenticated sender that a bot exists at this address and is merely turned off — an enumeration leak. Correct behaviour is total silence, identical to no gateway existing at all.

- **Telegram** (legacy path): removed the reply entirely; kept the admin-facing log warning.
- **Teams**: audited every other ingress path for the same class of leak and found a second, differently-shaped instance — `_gateway_settings()` never checked `is_enabled` at all, so a disabled gateway reached `ingest_gateway_event`, got rejected, and still returned HTTP 200 with a distinguishing message, while an unknown `gateway_name` hit an uncaught `DoesNotExistError` (404, no body). Fixed by raising the same exception for both, so the two states are now byte-identical to an external caller, not just quieter.
- Audited `gateway_webhook.py`, `slack_events.py`, `discord.py`, `automation_webhook.py`, and the email adapter — all already gate on `is_enabled`/`is_active` before any reply is possible, so no further instances exist.
- **Known gap, disclosed rather than hidden:** `teams_webhook.py`'s existing test suite mocks `_gateway_settings()` entirely, so this specific fix has no direct regression test covering it. The fix was verified by direct code reading and diff review, and the Telegram side's equivalent has a passing test on a live bench (`test_telegram_webhook_admission`, 3/3, re-run after this change).

Live-verified: bench restarted with the new code, backend admission tests re-run and pass (3/3), bench healthy.

PR #644, opened 2026-08-23
feat: Huf Developer Platform — public API v1 (Phases 0-3)

esafwan:
## Summary

Implements Phases 0–3 of the Huf Developer Platform spec (full spec: `Tracks/safwan-erooth.DeveloperSettings/GOAL.md` in the internal workspace repo, not part of this diff). Goal: a stable, versioned public API (`/huf/api/v1/...`) over Huf's existing agent runtime, plus developer-managed API credentials, so external developers can build against Huf without knowing its internal Frappe/DocType structure.

**This is a scoped subset of a 7-phase spec.** Phases 0–3 are implemented; Phases 4–7 (multimodal API, OpenAPI/SDKs/rate-limiting polish, OAuth Apps, Webhooks) are **not attempted** in this PR. See "Explicitly not done" below.

## What's in this PR

**Phase 0 — Audit** (not shipped as code; informs the design)
- Full inventory of ~130 existing whitelisted methods and 200+ frontend API call sites, synthesized into a capability matrix. Confirms every V1-scoped capability already has working runtime + an ad hoc API — this PR is adapter work, not new runtime.

**Phase 1 — API foundation** (`huf/api/v1/`)
- `/huf/api/v1/*` routing via a `BaseRenderer` (`router.py`), following the exact pattern of the existing `AgentStreamRenderer` — registered in `hooks.py` alongside it, no new routing mechanism invented.
- `RequestContext` (request id, user, auth mode), standard success/error JSON envelopes, `ApiError` hierarchy (400/401/403/404/429), an auth resolver chain (`resolve_principal`) designed so API-key/OAuth resolvers slot in without changing call sites, and a minimal OpenAPI scaffold.

**Phase 2 — Core agent API**
- `GET /v1/me`, `GET /v1/agents` (+`/{id}`), `GET/POST /v1/conversations` (+`/{id}`, `/{id}/messages`), `POST /v1/responses` (sync, queue-first — status is reported honestly as `queued`, no fabricated synchronous output), `POST /v1/responses?stream=true` (SSE — replicates the existing `run_immediately`-only streaming constraint and queued-runs ordering check exactly, does not weaken it), `GET /v1/runs/{id}`.
- Every handler is a thin wrapper reusing existing authorization primitives (`check_agent_access`, `assert_agent_access`, conversation-ownership checks) — no authorization logic is duplicated.

**Phase 3 — Credentials & Developer Settings**
- New `Huf API Key` DocType: SHA-256-hashed secret (rationale for not using a slow KDF documented in the code — these are high-entropy machine-generated tokens, not passwords), scopes (`agents:read`, `agents:run`, `conversations:read/write`, `files:read/write`, `voice:use`, `ocr:use`), agent restriction (all vs. selected), expiry, revocation. Raw secret is shown exactly once at creation and is never stored or retrievable again.
- `developer.access` / `developer.keys.manage` capabilities added to the existing Huf Role/capability system (no parallel permission system).
- API-key authentication (`Authorization: Bearer huf_sk_...`) resolves identity; **scope + agent-restriction enforcement is applied on every v1 endpoint** so a key can only ever reduce the owning user's authority, never increase it (`huf/api/v1/scopes.py`).
- **Settings → Developer** UI: Overview + a full API Keys section (create with scope/agent/expiry picker, list with status/scopes/last-used, revoke with confirmation, one-time secret reveal with copy-to-clipboard). Replaces an earlier placeholder page from this track.
- Partial: key usage updates `last_used_at`, but a real structured audit log (who/when/which endpoint) was not built — documented gap, not silently dropped.

## Explicitly not done (Phases 4–7, out of scope for this PR)
- Files/multimodal API, OCR extraction endpoint, voice session normalization under `/v1`
- Polished OpenAPI, interactive docs, TypeScript/Python SDKs, idempotency, pagination, rate limiting, usage/request logs
- OAuth Apps
- Webhooks

## How this was built / reviewed

Built with parallel sub-agents on tightly scoped, independent modules (one file per endpoint area), each given the actual existing runtime code to wrap rather than free rein to invent behavior. Every module was personally read and diff-reviewed before being wired into the shared `router.py` (the one coordination point, kept single-threaded to avoid merge conflicts across parallel agents). Full task-by-task log: `TRACKER.md` in the track directory (internal workspace repo).

## Verification status — please read before merging

**No live Frappe bench was available in the environment this was built in.** Verification was limited to:
- `ast.parse` syntax checks on every Python file
- `tsc -b --noEmit` (clean) on the frontend changes
- Manual review of every diff against the existing runtime it wraps

**Not verified end-to-end**: actual HTTP routing through `website_route_rules` (path-param extraction, route ordering), the SSE streaming path under Frappe's real async/event-loop request lifecycle, the API-key DB lookup/hash-compare under real data, and the full UI flow against a running backend. **A `bench migrate` + live smoke test of every endpoint in `TRACKER.md`'s final checklist is required before this is trustworthy in production.**

## Test plan
- [ ] `bench migrate` on a dev site to create the `Huf API Key` table
- [ ] `curl`/httpie smoke test: `GET /huf/api/v1/ping`, `GET /huf/api/v1/me` (session cookie), `GET /huf/api/v1/agents`, `POST /huf/api/v1/conversations`, `POST /huf/api/v1/responses`, `POST /huf/api/v1/responses?stream=true` against a `run_immediately` agent, `GET /huf/api/v1/runs/{id}`
- [ ] Create an API key via Settings → Developer, confirm the raw secret is shown once, then use it via `Authorization: Bearer` against a scoped endpoint and confirm both the allow and deny paths
- [ ] Revoke a key and confirm it stops authenticating
- [ ] `yarn typecheck` / `yarn lint` in `frontend/`

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -92,6 +92,7 @@ const HubSimplePage = lazy(() => import('./pages/HubSimplePage'));
 const GatewaysPage = lazy(() => import('./pages/GatewaysPage'));
 const AgentSettingsPage = lazy(() => import('./pages/AgentSettingsPage'));
 const GeneralSettingsPage = lazy(() => import('./pages/GeneralSettingsPage'));
+const DeveloperSettingsPage = lazy(() => import('./pages/DeveloperSettingsPage'));
 
 import { useEffect } from 'react';
 import { RouteErrorBoundary, clearChunkReloadFlag } from './components/RouteErrorBoundary';
@@ -569,6 +570,18 @@ function AppShell() {
                 <UnifiedLayout>
                   <Suspense fallback={<PageLoader />}>
                     <GeneralSettingsPage />
+                  </Suspense>
+                </UnifiedLayout>
+              </ProtectedRoute>
+            }
+          />
+          <Route
+            path="/settings/developer"
+            element={
+              <ProtectedRoute>
+                <UnifiedLayout>
+                  <Suspense fallback={<PageLoader />}>
+                    <DeveloperSettingsPage />
                   </Suspense>
                 </UnifiedLayout>
               </ProtectedRoute>
--- a/frontend/src/components/app-sidebar.tsx
+++ b/frontend/src/components/app-sidebar.tsx
@@ -159,6 +159,7 @@ export const settingsNavGroups: Array<{ label?: string; items: Array<{ title: st
         icon: Users,
         capability: ["users.manage", "roles.manage"],
       },
+      { title: "Developer", url: "/settings/developer", icon: Terminal, capability: null },
     ],
   },
 ]
--- a/frontend/src/components/settings/AgentApiDocs.tsx
+++ b/frontend/src/components/settings/AgentApiDocs.tsx
@@ -1,69 +1,128 @@
+import { useState } from 'react';
 import { Card, CardContent, CardHeader, CardTitle, CardDescription } from '@/components/ui/card';
 import { Badge } from '@/components/ui/badge';
+import { Tabs, TabsList, TabsTrigger } from '@/components/ui/tabs';
+import { ToggleGroup, ToggleGroupItem } from '@/components/ui/toggle-group';
 import { CodeBlock, CodeBlockCopyButton } from '@/components/ai-elements/code-block';
+import {
+  buildSnippet,
+  buildLoginSnippet,
+  type AuthMode,
+  type SnippetLanguage,
+  type SnippetRequest,
+} from '@/lib/apiSnippets';
 
 interface AgentApiDocsProps {
   baseUrl: string;
 }
 
 interface EndpointDoc {
-  method: string;
-  path: string;
+  request: SnippetRequest;
   description: string;
-  example: string;
 }
 
-function buildEndpoints(baseUrl: string): EndpointDoc[] {
-  return [
-    {
-      method: 'GET',
-      path: '/agents',
-      description: 'List every agent your key is permitted to use.',
-      example: `curl "${baseUrl}/agents" \\
-  -H "X-Huf-Api-Key: huf_sk_..."`,
-    },
-    {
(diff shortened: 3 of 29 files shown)


esafwan:
## Update: live-verified against a real bench

Provisioned a disposable bench off this exact branch (via `frappe-multihand`) and ran the whole stack live — curl smoke tests, real API-key create/auth/revoke, and the browser UI. This surfaced and fixed three real bugs static review couldn't catch:

1. **Fixture install failure** — `Huf Role Permission`'s `capability` Select field had a stale hardcoded options list; `bench migrate` failed inserting default role permissions. Fixed in `ae3aa6ec`.
2. **API-key auth was completely broken** — `Authorization: Bearer <key>` never reached our code at all; Frappe's own `frappe.auth.validate_auth()` intercepts any two-part `Authorization` header itself and hard-fails to "Session Expired" first. Moved to a dedicated `X-Huf-Api-Key` header. Fixed in `252adec3`.
3. **`/v1/me` reported "Guest" under API-key auth** — it called the session-bound `get_me()` instead of using the resolved principal. Fixed in `2dcb64dc`.

After these fixes, confirmed live:
- `/v1/ping`, `/v1/me`, `/v1/agents`, `/v1/conversations` — correct envelopes, correct 401s
- API key create → auth via `X-Huf-Api-Key` → correct identity (not Guest) → scope enforcement (`/agents` allowed, `/conversations` denied 403 for an agents-only-scoped key) → revoke → immediate 401
- Full browser flow: Settings → Developer renders, Create Key dialog works, one-time secret reveal + copy, list refreshes correctly

The "verification status" section of the PR description is now backed by an actual run, not just a caveat. Test-plan checkboxes there can be considered done for Phases 0-3's core paths; multimodal/OAuth/Webhooks (Phases 4-7) remain out of scope for this PR.

PR #645, opened 2026-08-23
analytics: fix token accounting, then build Run → Conversation → Agent → Model → Provider views

esafwan:
Implements the audit and target design from
`claude/huf-analytics-observability-audit-mz29re`
(`docs/analytics-observability-architecture.md`), all six phases.

The audit's own framing was **repair and extend, not rebuild** — the foundation here
(`context_segments.py`, the scheduled rollup, the cost calculator's priority chain) is sound. But it
surfaced defects serious enough that building conversation/agent/model dashboards on today's numbers
would produce *confidently wrong* charts. This PR fixes those first, then builds on top.

---

## The defect that mattered most

`input_tokens` meant **two different things depending on which code path ran**.

- Sync (`run()`) accumulated `prompt_tokens` across every tool-calling round — correct.
- Streaming (`run_stream()`) reset `stream_usage = None` at the top of each round and never
  accumulated, so after the loop it held **only the last round's** prompt count.

A five-round streaming run reported the size of its final prompt; an identical sync run reported the
true sum. Any aggregate mixing both — every existing dashboard — was incorrect at the source. The same
bug affected streaming **cost**, computed once post-loop from the last round while sync accumulated per
round.

Both paths now accumulate per round and return one contract. `peak_context_tokens` is tracked
separately as a **max, never a sum**, because the two answer different questions:

| Metric | Answers |
|---|---|
| `billed_input_tokens` | "What was I charged for?" |
| `peak_context_tokens` | "How close did this run get to the context window?" |

Conflating them is the misleading aggregation this whole project exists to avoid, so they are kept
structurally apart all the way into the UI.

---

## Defect coverage

| ID | Defect | Where fixed |
|---|---|---|
| D1 | `input_tokens` path-dependent | `providers/litellm.py` — both loops accumulate; peak tracked as max |
| D2 | `segment_tokens` is round-1-only, tool exchange unmeasured | `context_segments.count_tool_exchange_tokens` |
| D3 | `cache_miss_tokens` a mislabelled duplicate of cache writes | removed from provider dicts and both snapshot writes |
| D4 | tool-kind runs had NULL `conversation` | `flow_engine._create_flow_agent_run` |
| D5 | usage extraction duplicated 4× and drifted | new `huf/ai/usage_extraction.py` |
| D6 | legacy fallback drops cache/cost silently | `provider_path` recorded — but see *Finding* below |
| D7 | cache/total/skip only inside JSON | promoted to flat `Agent Run` columns |
| D8 | rollup has no `conversation` dimension | `DIMENSION_FIELDS`, appended last |
| D9 | context window hardcoded 200K | `model_metadata.resolve_model_context_window`; `None` when unknown |
| D10 | `tools` one undifferentiated count | `tools_breakdown` by source and per tool |
| D11 | 10,000-row client-side reduction | `HomePage` on the aggregate API |
| D12 | `series`/`breakdowns` never rendered | new `/analytics` page + `dimension` param |
| D13 | unbacked cache stats | **not fixed** — now labelled as indicators, not measurements; see below |
| D14 | no cache-write pricing | `cost_calculator` + new AI Model field |
| D15 | assembled system prompt never persisted | **not implemented** — see below |
| D16 | no composition reconciliation | `context_segments.reconcile_composition` |

**Two plan items are deliberately not implemented, and are called out rather than quietly dropped:**

- **D13** — `wasted_writes_tokens` stays hardcoded `None` and `prefix_stability` stays a hash
  comparison. A real wasted-writes figure needs read-after-write tracking no provider reports today,
  and confirming prefix stability against actual cache behaviour needs per-breakpoint attribution that
  `context_segments.py` deliberately declines to guess at. What this PR *does* fix is the presentation:
  these are now labelled as indicators, not measurements.
- **D15** — the assembled system prompt is still not persisted. Storing it for every run is a
  materially different change: unbounded storage growth plus a new PII surface, since assembled prompts
  contain memory blocks and injected knowledge. It deserves its own retention/redaction decision rather
  than being smuggled into an analytics PR.

---

## Decisions where the plan was ambiguous or self-contradictory

**The plan contradicted itself on tool sub-typing.** §4.2 asks for `tools: {total, by_source}` while
also requiring the five `segment_tokens` keys stay stable. Both cannot hold: `context-bar.tsx` reads
`segments.tools` as a number and multiplies it into a width percentage, so making it an object would
silently break the composition bar and every historical snapshot. Resolved by keeping `tools` an int
and putting the sub-typing in a sibling `tools_breakdown` key.

**The plan's migration note, taken literally, would have preserved a known-wrong column.** §8 floats
leaving `input_tokens` "frozen as legacy", which would mean deliberately continuing to write the wrong
last-round value on the streaming path — contradicting §4.1's own mandate for path-independent numbers.
Instead: `input_tokens` is fixed, the explicitly-named canonical columns are added alongside, no
historical row is rewritten, and the rollup reads `COALESCE(billed_input_tokens, input_tokens)`.

**Charts (§7.2) — Option B, scoped to exactly one.** No first-party page in this product renders a
chart today; recharts appears only inside the sandboxed AI-artifact renderer. Context growth over turns
is the one question tiles genuinely cannot answer, so it gets the one chart. Everything else is built
from existing `GaugeRow` / `MetricGauge` / `ContextBar` primitives.

---

## Unknown is not zero

`None` means "could not be counted"; `0` means "counted, empty". A silent `0` makes every composition
percentage quietly wrong, so unknowns propagate as unknown end to end — through
`_accumulate_composition` (a segment total becomes and stays `None` once any contributing run reports
`None`), through the API, into TypeScript types with `number | null` **values**, and into components
that render "not measured".

Concretely, this is why:
- an unknown context window draws **no headroom bar** and captions itself, rather than inventing a
  denominator;
- unmeasured turns **break** the context-growth line (`connectNulls={false}`) rather than plotting as 0;
- the `/analytics` composition view **excludes** unmeasured segments from the denominator instead of
  letting them silently shrink the others.

## Honest disclosure over quiet under-reporting

- Historical `run_kind="tool"` runs have NULL `conversation` and cannot be retro-linked. The
  conversation pane says so explicitly rather than under-reporting silently.
- The breakdown table shows "Top N of M" — a silent cap reads as "this is everything".
- Runs served by the legacy fallback are flagged as having incomplete cache and cost data.
- The backfill patch promotes only the three values that genuinely exist in historical
  `usage_snapshot`. The other six columns stay NULL, and the patch docstring says why they must never
  be zero-filled.

---

## A finding that corrects the audit

The audit describes D6 as an invisible data-quality cliff blended into aggregates. In practice it is
**near-dead**: `RunProvider.run()` is a *sync* method returning `litellm.run(...)`'s unawaited
coroutine, so its `try/except Exception` wraps only coroutine *creation* — the legacy fallback catches
essentially nothing but `ImportError` and never fires for runtime LLM errors. `provider_path` is still
worth recording (it makes the assumption checkable rather than assumed), but this PR does not claim to
have closed an active hole.

That same async shape is worth flagging for reviewers: the obvious way to tag the result
(`result.provider_path = ...`) raises `AttributeError`, because coroutines have no `__dict__`. Hence
the `_await_tagged` wrapper.

---

## Testing

75 tests pass, including a D1 regression that replays a 3-round `[100, 250, 180]` workload and asserts
sync- and stream-shaped accumulators converge on `billed=530, peak=250` — with a canary proving the old
reset-per-round behaviour would report `180`, so the test can actually fail if the bug returns.

Frontend `yarn typecheck` clean; `yarn build` succeeds.

`huf/ai/tests/_stub_env.py` is added because `frappe` submodule imports need real stubs, letting these
suites run standalone outside a bench.

---

## Migration

- One patch backfills `cache_creation_tokens`, `total_tokens` and `cache_skipped_unsupported_model`
  from existing `usage_snapshot` JSON. Idempotent, batched, safe on malformed rows.
- `usage_snapshot` goes to `schema_version: 2`, additively. Verified nothing reads it expecting 1.
- Adding the `conversation` rollup dimension needs one `refresh_rollups(full_backfill=True)`; the
  mechanism already exists and is already triggered lazily on a cold window.
- `conversation` is appended **last** in `DIMENSION_FIELDS` so previously-stored `dimension_key`
  strings stay positionally decodable.
- All three doctype diffs are purely additive; no existing field is reordered or altered.

## Independent review

A separate reviewer verified this against the plan defect-by-defect and found **no severe defects**. It
independently confirmed the D1 fix reaches the accumulation on every round-completing control-flow path
(tool-call break, `is_stop` break, plain exhaustion) while error paths return untouched, traced the
unknown-vs-zero discipline end to end without finding a coercion point, and re-ran the suites (75 pass,
typecheck clean, all three doctype `field_order` arrays intact).

Its one substantive finding — a latent `NameError` if the conversation-metrics update failed before
`total_tokens` was assigned, since that `except` swallows and a later write reads the variable — is
fixed in this branch by hoisting the assignment out of the swallowing `try`.

## Review notes

- The `Agent Run` and `AI Model` schema additions are the parts worth the closest look, since they are
  the hardest to change later.
- The streaming accumulation point in `run_stream()` is placed after the chunk loop and before
  `if is_stop: break`, so it is reached on every round-completing path while error paths still return
  untouched. Worth verifying independently.


esafwan:
## Live bench verification

Provisioned a disposable bench on this branch (blank site, `bench install-app huf`) to check the parts
unit tests cannot reach — that the schema actually migrates and the endpoints actually respond.

**All 15 new columns exist in the database after install:**

| Doctype | Columns |
|---|---|
| `Agent Run` | `billed_input_tokens`, `peak_context_tokens`, `cache_creation_tokens`, `total_tokens`, `cache_skipped_unsupported_model`, `round_count`, `model_context_window`, `provider_path`, `execution_mode` |
| `AI Model` | `context_window`, `max_output_tokens`, `cached_input_write_cost_per_1m_tokens` |
| `Agent Run Analytics Rollup` | `conversation`, `cache_creation_tokens`, `composition_totals` |

**Runtime checks:**

- `backfill_agent_run_usage_columns.execute()` runs clean, including a deliberate second run — it is
  idempotent as documented.
- `refresh_rollups(full_backfill=True)` completes with the widened dimension tuple
  `('agent', 'provider', 'model', 'run_kind', 'conversation')` — `conversation` last, so stored
  `dimension_key` strings stay decodable.
- `get_execution_analytics` responds for **all five** dimensions (`provider`, `agent`, `model`,
  `conversation`, `run_kind`), each returning the full shape and echoing `metadata.dimension`.
- `get_conversation_analytics` on a nonexistent conversation raises `DoesNotExistError` cleanly rather
  than a 500.
- `reconcile_composition` unknown-propagation confirmed against real code: an uncountable tool exchange
  returns `None`, a `None` segment returns `None`, and a genuine in-tolerance case reconciles
  (`delta_ratio: 0.0, within_tolerance: True`).

**Not included: UI screenshots.** The bench is a blank site with no Agent Runs, so the analytics
surfaces would only show their empty states — which would illustrate the `EmptyState` handling but not
the charts or composition bars. Producing meaningful screenshots needs a site with real run history
against a configured LLM provider. Flagging that honestly rather than attaching screenshots of empty
pages and implying they show the feature working.

esafwan:
## Follow-up to review: 3 more commits

Acting on review discussion — two things the review flagged, plus a design refinement to how D13/D15
are handled.

**1. Dead ternaries removed** (`a0339e6f`). The two `cached_tokens if usage else None` /
`cache_creation_tokens if usage else None` lines the review flagged were unreachable — both sit inside
a block already guarded by an enclosing `if usage:`. Removed; no behavior change, 58/58 relevant tests
still pass.

**2. `wasted_writes_tokens` removed rather than left as a permanent `None`** (`e5dd5138`). On reflection
this is not the same kind of gap as `prefix_stability`. It was hardcoded `None` everywhere with a
comment ("needs read-after-write tracking; not captured yet") and no path to ever becoming real without
a new cross-run correlation subsystem — every run on every site was rendering a UI tile that could only
ever say "Not yet tracked". Advertising a metric that will never exist is worse than not having the
field, so it's gone: from `cache_metrics.py`'s return dict, the TS type, and the run-detail UI (4-col
grid → 3-col). `prefix_stability` is untouched — it's a real, if approximate, computed value (hash
comparison across consecutive runs), already labeled as an indicator rather than a measurement, so it
stays.

**3. D15 gets an opt-in escape hatch** (`53e37c20`). The PR still does not persist the assembled system
prompt by default — that's still the right default given the storage growth and PII surface. But some
deployments have the opposite obligation: a legal/compliance need to retain the exact prompt a run saw,
sometimes with a rotation policy. Added a system-wide, off-by-default switch:

```
bench set-config huf_retain_system_prompts_enabled true
```

When set, `huf/ai/system_prompt_retention.py` writes the verbatim assembled instructions to a new
**Agent Run Prompt Snapshot** doc — deliberately not onto `Agent Run` itself, so a future retention/
rotation job can purge by `captured_at` without touching the Agent Run table, and so the snapshot can
carry its own, tighter permissions (System Manager / Huf Manager, read-only; writes are system-created
only). Wired at both persistence sites, reading the same `agent.instructions` text
`context_segments.py` already counts the size of. Best-effort — a snapshot failure is logged and never
fails the run.

**Not built here, on purpose:** the rotation/purge job itself, and a per-agent override that would
further narrow this site-wide switch once there's a field on the Agent doctype for it (this PR provides
the system-level gate that override would additionally need to check).

5 new standalone tests cover the retention helper (disabled-by-default, verbatim write when enabled,
no-instructions no-op, insert-failure-is-swallowed, site-config read). 63/63 standalone-safe tests pass;
frontend typecheck and full production build both clean.

esafwan:
## Live verification of the 3 follow-up commits

Provisioned a second disposable bench (\`analytics-obs2\`) on the updated branch and verified against
a real database, not just standalone unit tests:

- \`Agent Run Prompt Snapshot\` doctype created cleanly by \`install-app\`, all 5 fields present
  (\`agent_run\`, \`agent\`, \`conversation\`, \`captured_at\`, \`system_prompt\`).
- \`wasted_writes_tokens\` confirmed absent from \`compute_run_metrics()\`'s return — live call returns
  exactly \`['cache_read_share', 'counterfactual_savings', 'effective_input_multiplier',
  'prefix_stability']\`.
- \`maybe_snapshot_system_prompt\` exercised against a real \`Agent Run\` row and a real site_config:
  disabled (default) → no snapshot row created; enabled → exactly one row created, with
  \`agent_run\`/\`system_prompt\` matching what was passed in verbatim.

Bench torn down after verification; nothing left running.

esafwan:
## Final review before merge

Ran an independent, adversarial final review over the full branch (all 10 commits, including the 4
follow-ups from review discussion, which had only been self-verified until now). It re-traced the D1
control flow line-by-line rather than trusting the docstring, independently re-checked each of the 4
follow-up commits' claims against the actual diff, ran all 83 standalone-runnable tests, and confirmed
a clean frontend typecheck.

**Verdict: no blocking issues. MERGE.**

One non-blocking note carried forward: `get_execution_analytics` can trigger a synchronous
`refresh_rollups(full_backfill=True)` from a GET request when the rollup table is cold for the
requested window — pre-existing design from the original 6 commits, not introduced by the follow-ups,
worth a should-fix-later look but not a merge blocker.

Merging to `pre-dev`.

PR #646, opened 2026-08-23
feat: deterministic Agent Procedure execution, on a shared IR with Flow

esafwan:
## What this is

A deterministic execution path for Agents. Today a stable, repeated procedure costs one model turn per
tool call — the model re-derives which tool comes next, which result feeds which argument, and what to
extract, every single time. This adds **Agent Procedure**: a versioned, statically-inspectable graph the
model selects once and supplies typed inputs to, which then runs to completion without further model
involvement and returns a bounded result.

The design principle, from the originating plan: **deterministic by default where the procedure is known,
agentic by exception where judgment is actually required.**

Flow and Agent Procedure now share one IR and one executor. Flow is what you draw; a Procedure is what it
compiles to.

## Why it also repairs Flow and Context Artifacts

Six of Flow's defects were things the Procedure runtime had to build correctly anyway, so building them
twice would have been the expensive option:

| Defect | Fixed by |
|---|---|
| Version auto-incremented on every save; a resumed run re-fetched the *current* definition | Immutable content-addressed versions, pinned by the run |
| Four divergent `{{...}}` interpolations with different semantics per node type | One structured `{"$from": "node.path"}` reference |
| `loop` burned one hop per iteration and died at `max_hops` 100 | Bounded `foreach` with its own budget |
| No lock on `run_flow`; two concurrent calls advanced the same cursor | Distributed run lock |
| Validator never checked node config, reachability or cycles | Profile-driven static validator |
| Context re-parsed at ~15 call sites | One context object |
| `Flow Definition.is_system` was a dead field | Shared system-record locking |

`Agent Context Artifact` followed the same pattern. It was ~15% built: the drill-down returned an
always-empty field, ordinary users could not read artifacts from their own conversations, there was no
generic producer, nothing ever told the model an artifact existed, and the store had no lifecycle or
quotas. All fixed here.

## Security fix

**A Flow `tool.call` node could bypass all permission checks.** `flow_tool_executor` had no permission
machinery and passed caller args straight to handlers, which accept `ignore_permissions=False, **kwargs` —
so a caller-supplied `ignore_permissions=True` reached them. The LLM path strips exactly this; the Flow
path did not. Separately, 15 built-in tools (including `delete_document`, `submit_document`,
`cancel_document`, `set_value`) resolved through an alias map with no `Agent Tool Function` record and no
registry gating. `Flow Definition` is writable by **Huf Manager**, and flows are agent-triggerable.

Not exploitable in the field — zero Flow Definitions exist anywhere — but real. Closed structurally: one
invocation service now owns permission checks, the `ignore_permissions` strip, and telemetry, so no caller
can skip them.

## Invariant status

An independent adversarial audit ran against all commits (`INVARIANT-AUDIT.md`).

| | Status |
|---|---|
| I1 no authority escalation · I2 runtime enforcement always runs · I3 static tool closure · I4 no LLM in the path · I5 telemetry never collapsed · I7 bounded output, fail closed | **Hold** |
| I6 immutable versions | **Hold for identity** (the docname is `procedure_id-vN`, so the primary key rejects a colliding insert) + **tamper-evident for content**: `verify_fingerprint` runs before execution, so a `db.set_value`/raw-SQL rewrite is detected and fails closed |
| I8 no automatic write activation | **Hold** — only read-only procedures may be bound, enforced at save *and* re-checked at read |
| I9 failure degrades, never breaks the agent | **Hold** — every path returns a branchable result rather than raising |
| I10 N-run validation | **Mechanism exists, not yet exercised against a live agent** — see limitations |

The audit found the fallback protocol was fully built but never called. That is now wired in.

## Test evidence

Baseline on `pre-dev` before any change: **901 tests, 3 failures, 75 errors, 144 skipped** — the suite was
never green. Every gate in this work asserted *no new failures against that recorded list*, not "green".

**445 tests added by this branch, all passing module-by-module on a real bench** (18 modules; full list in
the tracker). Frontend: `yarn typecheck`, `yarn build` and 148 vitest tests pass.

**Aggregate caveat, stated plainly.** Running the whole suite in one process shows more failures than the
baseline. This is pre-existing cross-test state leakage that adding ~500 document-inserting tests exposes,
not a product regression: every affected module passes in isolation, and the failures are
`frappe.flags.currently_saving` being nulled and `frappe.session.user` not being restored by tests this
branch does not touch. This PR **fixes one such leak** — `test_faiss_backend` replaced `frappe.local.flags`
and never restored it, which alone accounted for **38 of the baseline's own errors**. The remaining leaks
are pre-existing and out of scope here; fixing them belongs in a dedicated test-infrastructure change.

## Update — 2026-08-24: schema migration + Procedure UI, live-verified

Everything below was previously a limitation. Since the initial draft:

- **Gap closed: Flow's stored `definition_json` now uses the shared IR.** `flow_definition.py`'s
  `_validate_definition_json` delegates to `huf.ai.graph.validator.validate_flow_graph` — the same gate
  `analyze_conversion` runs before Flow→Procedure conversion. Nodes carry their own `next`/`on_error`
  (no top-level `edges` array), `schema_version` is `"1.0.0"`. `flow_engine.py` derives routing from
  `next`/`on_error` at runtime, with a compatibility path (`effective_edges()`) so any `Flow Run` pinned
  under the old shape still finishes. Frontend save path (`flowSerializer.ts`, `flowApi.ts`) updated to
  match. A Flow built through the real builder can now actually convert.
- **Gap closed: an "Agent Procedure" authoring/browsing UI exists.** `/procedures` (list, with tier/status
  filters), `/procedures/:id` (detail — definition JSON, fingerprint, validation status, version history),
  and a "Procedures" tab on the Agent form for managing `Agent Procedure Binding` records (picker filtered
  to `is_read_only` procedures only — write-tier procedures are correctly excluded, since binding a write
  procedure has no approval surface yet, see remaining limitations).
- **Gap closed: the T-50 validation harness is now reachable.** A "Test" button on the Procedure detail
  page calls a new whitelisted `run_validation_harness` endpoint and renders the promotion decision inline.
- **Test suite: zero regressions.** Full `bench run-tests --app huf` on bench `detproc-fbb1`:
  1410 tests, **3 failures / 75 errors** — identical counts to `BASELINE.md`, confirmed line-by-line to be
  pre-existing (`huf_role`, `integration_service`, `gateway_service`, `queue_first_runs`), none touching
  this change. `bench migrate` on the merged branch completed with no errors.
- **Live-verified, not just reviewed.** All of the above was clicked through in a real browser against a
  real bench (list page showing real Draft procedures, detail page, Test button producing a real harness
  result, binding tab showing an existing binding and successfully listing eligible read-only procedures
  in its picker) — not static code review. Screenshots available on request.

Still open, not attempted in this pass:
- **"View graph" is not functional for procedures.** Still routes to `/flows/:flowId`, which renders
  `Flow Definition` only — a Procedure-aware graph viewer doesn't exist.
- **A minor, likely pre-existing UI issue**: a "Failed to load Procedures" toast fires on the Agent
  form's Procedures tab even though the binding list and picker both load correctly. Not reproduced in the
  new files added by this batch; not chased down further.
- **No production path for write Procedures.** I8 correctly blocks binding them; there is still no
  review/approval UI for that path (deliberately out of scope for this batch).

## Limitations — please read before approving

1. **Benchmark tool layers are simulated.** Benchmarks 1, 3 and 4 execute end to end and pass their
   invariants, but against hand-written fake invokers. This proves the runtime threads scope filters into
   query arguments, enforces bounds, stays deterministic under concurrency, and does not duplicate writes.
   It does **not** prove any tool is wired to real ERPNext.
2. **No measured token or latency reduction.** The validation harness exists and gates promotion, but its
   headline numbers are illustrative, not measurements against a live agent. The feature's central claim —
   that this is cheaper — is **not yet evidenced**. That measurement is the first thing to do next.

## Scope deliberately excluded

The trace-mining compiler (learning procedures from observed runs), lazy *recipe* generation, sandboxed
code transforms, and the circuit breaker. `T-52` converts an existing Flow; it does not mine trajectories.

## Review order suggested

1. The security fix — `huf/ai/tool_invocation.py`, and that `flow_tool_executor` now delegates.
2. `huf/ai/graph/executor.py` and `procedure_runtime.py` — the core.
3. `AgentProcedure.autoname` — version immutability rests on it.
4. `INVARIANT-AUDIT.md` and the limitations above.


esafwan:
## Live verification — 2026-08-24

Ran the full punch list from the last review against bench `detproc-fbb1` (real browser, real data, not static review):

- `bench migrate` on the merged branch — clean, no errors.
- `bench run-tests --app huf` (full suite, 1410 tests) — **3 failures / 75 errors**, identical to `BASELINE.md`. Confirmed none of them touch `flow_definition`, `flow_engine`, or `procedure_validation_api`. One real bug was found and fixed in this pass: the auto-convert checkbox tests still built Flows with the pre-migration `id`/`edges` shape, which the new stricter schema validation now rejects — fixed by switching the test fixture to a schema-valid shared-IR definition (see the two `fix(test)` commits).
- Frontend `tsc -b` and a full `vite build` — clean, and produced the new page chunks (`AgentProceduresPage`, `AgentProcedureDetailPageWrapper`, `agentProcedureApi`).
- Clicked through in a browser with a real Administrator session:
  - `/procedures` — list renders real Draft procedures with tier/status filters.
  - `/procedures/:id` — detail page renders fingerprint, schema version, read-only/write-contains flags, validation status.
  - **Test button** — calls the new `run_validation_harness` endpoint live and renders a real promotion decision inline ("Not approved for promotion" with reasons, correctly, for a minimal test procedure).
  - Agent form → More → Procedures tab — shows the existing binding, and the "bind a new procedure" picker correctly lists only `is_read_only` procedures.

One thing to note, not chased down: a "Failed to load Procedures" toast fires on the Agent's Procedures tab even though the binding list and picker both load and work correctly — likely a background poll unrelated to the new components (not reproduced in source of the files this batch added). Left as a known minor issue rather than guessing at a fix under time pressure.

PR description updated with the same detail. Full state, evidence, and screenshots described in [`GAP-ANALYSIS-ADOPTION.md`](https://github.com/tridz-dev/huf/blob/feat/agent-procedures/Tracks/safwan-erooth.DeterministicAgent/GAP-ANALYSIS-ADOPTION.md) — note that file lives under `Tracks/`, which is gitignored, so it's not actually in this diff; happy to paste its content into a PR comment if useful for reviewers who don't have the track checkout.

esafwan:
## Follow-up — fixed the two loose ends from the previous verification pass

- **"Failed to load Procedures" toast**: root-caused. `AgentFormPage`'s generic per-section lazy loader was firing for the new `procedures` tab, which has no corresponding Agent doctype section — `ProcedureBindingsTab` fetches its own data independently. Added `procedures` to the tabs that skip that generic fetch. Verified live: the toast no longer appears when opening the tab, and the binding list/picker still work correctly.
- **"View graph" broken link**: fixed. It previously always routed to `/flows/:id` regardless of whether the run was Flow- or Procedure-backed. Now routes to `/procedures/:id` (which exists as of this batch) when the run carries a `procedureId`, falling back to the Flow route otherwise. Verified by code review + a clean `tsc -b` on the bench — not click-through verified, since no `Agent Procedure Run` exists yet on this bench to click from (would require triggering a real model-backed chat execution).

Re-ran the full suite after these two fixes: still 1410 tests, **3 failures / 75 errors**, identical to baseline.

`TRACKER.md` updated with the full session log entry for this batch.

esafwan:
## Gap 6 (write-procedure approval) + Gap 9 (ERPNext guided setup) — merged and live-verified

Both closed a real gap: I8 correctly forbids write procedures from auto-activating, but there was no path at all — not even a manual one — for a write procedure to ever become usable. Same for ERPNext master-data setup: real, by-hand work with no guided path.

**What shipped:**
- `Agent Procedure` gains `approval_status`/`approved_by`/`approved_at`/`approval_note`. Two new whitelisted endpoints (`request_procedure_approval`, `approve_procedure`, role-gated to System Manager/Huf Manager) in `huf/ai/procedure_approval_api.py`.
- `Agent Procedure Binding`'s guard now allows binding when `is_read_only=1` (unchanged default) **or** `is_read_only=0 and approval_status=="Approved"` — additive, not a loosening of the existing check.
- An "Approval" card on the Procedure detail page, visible only for write-tier procedures, with Request Review / Approve / Reject.
- `huf/ai/erpnext_demo_setup.py::ensure_erpnext_demo_masters()` — idempotent, creates the minimum ERPNext master data (Warehouse Type, Item/Customer Group, Territory, Price List, Fiscal Year) needed to exercise a realistic customer/invoice scenario. A "Set up ERPNext demo data" button under Settings → General → Developer tools (System-Manager-only).

**Live verification, not just static review:**
- `bench migrate` clean. Full suite: 1423 tests, 3 failures / 75 errors — identical to baseline (one new error surfaced on the first pass from a test not initializing `frappe.flags.currently_saving`, same pre-existing pattern hit elsewhere in this track; fixed and reverified back to exactly baseline).
- Clicked "Set up ERPNext demo data" live — created the 5 expected records, second click confirmed idempotency (`created: []`, everything `already_present`).
- Created a real write-tier test procedure (genuine write permission envelope, `is_read_only=0`), confirmed the Approval card renders with the correct copy, clicked Approve live, confirmed status flips to Approved with attribution, then confirmed via console that binding **succeeds only after** approval — this is the actual security-relevant behavior, verified end to end, not just read from source.

esafwan:
## New capability: "Run & Propose Procedure" — the actual gap this track was missing

Everything so far required a user to draw a Flow first. That's backwards — this closes the real gap: **a user just uses the agent normally, and after a completed turn, can explicitly ask Huf to compile that exact run into a proposed Procedure**, review it, and accept or decline. Nothing is saved until they click Accept.

### Design
- `huf/ai/procedure_proposal.py`: `propose_procedure_from_run(agent_run_name)` — a preview, saves nothing — compiles the run's real `Agent Tool Call` trace into a graph-IR Procedure. Value binding follows a strict priority: (a) a value traceable to the user's original prompt → becomes a named input parameter; (b) a value matching an earlier step's output → becomes a `$from` reference; (c) a small set of genuinely trivial constants → kept literal; (d) **anything else refuses the whole proposal** with a specific reason naming the step/tool/argument — never silently bakes in an unexplained (judgment-derived) value as a fake-deterministic literal. `accept_procedure_proposal(...)` re-validates the graph server-side before saving as Draft (never trusts a client-round-tripped graph).
- A dedicated **opus-5 medium-effort UI/UX review pass** ran against this feature specifically (separate from the sonnet implementation agents) and caught a real functional bug before it reached anyone: the review dialog was reading a graph shape (`steps`/`args`) that never matched what the backend actually returns (`nodes`/`config.input`/`$from`), so it would have rendered zero steps for every proposal. Fixed, plus a full copy pass — no more "this run"/"compiled steps"/raw tool ids, refusals framed as a normal outcome (not an error), a load-failure retry state, and a clear source-run link so a proposal can't get lost in a long conversation.

### Live verification
- `bench migrate` clean, full suite (1439 tests) **3 failures / 75 errors — identical to baseline.**
- `tsc -b` clean on the bench.
- Used a real Google/Gemini provider (`gemini-3.6-flash`, already configured on this bench) through the actual chat UI, not simulated:
  - Confirmed the D8 provenance row and this new "Save these steps as a procedure" button both render live on a completed real chat turn.
  - Clicked it against a run whose only step was itself a bound Procedure call — got the **correct refusal**, live, with the exact reviewed copy ("These steps can't be saved as a procedure... Nothing was saved or changed"), a working source-run link, and a **Close** (not Cancel) button.
  - Called `propose_procedure_from_run` directly against real production trace data via console — confirmed it returns the same structured refusal server-side.
- **Not click-verified**: the Accept path. This agent's current tool configuration (lazy tool discovery gates `list_notes`/`list_todos` behind the bound procedure) means every real ad-hoc LLM trace on this bench starts with a procedure call and correctly refuses — there's currently no real trace on this bench that *would* be accepted. The Accept path itself is unit-tested (16/16 passing, including a tampered-graph rejection test) and code-reviewed (confirmed server-side re-validation is real, not decorative) — but I'm not going to claim it as click-verified when it isn't. Worth testing against an agent with broader ad-hoc tool access before this is called fully proven.

### Explicit scope note
This is **not** GOAL.md's full "Analyse Agent" vision — that would mine patterns across *many* runs and proactively suggest a procedure unprompted. This compiles *one* run the user explicitly chose, on request. Naming that distinction plainly so it isn't overclaimed later.

esafwan:
## UX clarity pass, plus one serious pre-existing bug found and fixed

Prompted by real feedback while manually reviewing the Procedures UI: it was too engineer-flavored, gave zero explanation of what a "Procedure" is, and one displayed procedure name was literally the string **"tampered directly"** — which turned out to be real production data corruption (not test pollution): two of this session's own benchmark procedures (`team-digest-v1`, `customer-financial-context-v1`) had been overwritten via `frappe.db.set_value` during earlier manual console probing in this same session and never reverted. Fixed the data directly; both now show correct names.

### Explainer copy (docs-only, no logic changes)
- Procedures list page: plain-language one-liner ("A Procedure is a sequence of steps an agent already worked out in chat, saved so it can repeat them the same way every time...") plus a proper empty state explaining the chat→save flow, not diagram-drawing.
- Procedure detail page: a one-line gloss under the tier badges explaining what Draft/Read-only/Write actually mean for a non-engineer.
- Agent binding tab: rewrote from "Deterministic Agent Procedures this agent can execute..." to plain language about what binding does and why.

### Save-time discovery (closes gap 3 from the earlier goal-check)
Saving a Flow with "create a procedure on save" checked previously wrote a `conversion_note` field nobody ever saw. Now:
- The checkbox itself is wired into the Flow Settings modal (it existed only in the DocType schema before — no UI control to toggle it at all).
- After save, a toast surfaces the outcome — success or refusal — using the existing `conversion_note` text, framed as a normal outcome either way.

### A real, serious, pre-existing bug found and fixed in the same pass
While live-verifying the above, discovered that **every single Flow creation was broken** — clicking "New flow" always failed schema validation. Root cause: the shared graph-IR schema requires node ids matching `^[a-zA-Z_][a-zA-Z0-9_]*$` (no hyphens), but the default new-flow template (`empty-trigger`) and every canvas-added node (`node-<timestamp>`, `node-trigger-<timestamp>`) used hyphenated ids. This has nothing to do with this session's other work — it's a pre-existing defect from whenever the Flow builder was last touched — but it's completely blocking, so fixed here: switched to underscore-based ids throughout. Verified directly against the real schema validator (confirmed the old ids fail, confirmed the new ones pass), then live-verified end to end in a browser: clicked "New flow," got a real Draft Flow with 1 node, opened its canvas successfully. This was previously impossible.

### Verification
- `tsc -b`: clean.
- Full suite: 1439 tests, 3 failures / 75 errors — identical to baseline, across every round of this fix batch.
- Live: Procedures list + detail explainer copy confirmed rendering in a real browser. New-flow creation confirmed end-to-end (previously 100% broken, now works). Flow Settings modal checkbox/toast: code-reviewed and type-checked but not click-verified — this app's Settings/More dropdown modals were flaky to drive via this session's browser-automation harness all day (likely a portal-timing quirk in the automation tooling, not a real user-facing issue, but flagging honestly since I can't personally vouch for it being pixel-verified).

PR #647, opened 2026-08-23
feat: render ERPNext/Frappe list/form/report views inline in chat as artifacts

esafwan:
## Summary
Adds a generic, doctype-agnostic Frappe data surface for the chat agent, and renders it inline in chat as a new artifact family (`frappe-list` / `frappe-form` / `frappe-report`) — Claude-Artifact-style — with an "Open in Desk" deep link on each view. Built for huf's own deployment model: huf is a Frappe app installed on the same site as Desk/ERPNext, so `/app/...` links inherit the session with no SSO handoff needed (verified against `vite.config.ts`/`proxyOptions.ts`/hooks.py — see track PLAN.md "Phase 4 finding").

Per an architecture review, this does **not** add `frappe-ui-react` as a dependency — it's a community port, not Frappe-maintained, and would add a second styling/data-fetching stack. Instead it's built on the frontend's existing design-system primitives, `@tanstack/react-table`, and `frappe-js-sdk`.

## Backend (`huf/ai/tools/frappe_generic.py`, new)
- `handle_get_doctype_meta` — field metadata (types, Select options, Link targets, Table children, mandatory/read-only/depends_on) via `frappe.get_meta`.
- `handle_list_records` — generic `frappe.get_list` with filters/fields/pagination/order_by, returns `total_count`.
- `handle_get_record` — single record fetch.
- `handle_create_record` / `handle_update_record` — **draft-only**. These never write to the database; they validate and return a draft payload. The actual write happens client-side, as the logged-in user, through the normal Frappe REST path — never server-side on the agent's say-so.
- `handle_render_frappe_view` — composes the above into one of the three new artifact types.
- `erpnext_reports.handle_run_report` extended with `limit_start`/`limit_page_length` pagination (report API doesn't expose an unpaginated count, so `total_count` is `null` there by design).

### Security model (non-negotiable, enforced in every handler)
- Doctype denylist: `User`, `Role`, `Role Profile`, `Custom Script`, `Server Script`, `Property Setter`, `File`, any `OAuth*`/`Integration*` prefix, and all Single doctypes — deny-by-default if a doctype doesn't exist or its meta can't load.
- `frappe.has_permission(doctype, ptype=...)` enforced per operation.
- Reads go through `frappe.get_list`/`frappe.get_doc` (permission-query conditions applied), never raw SQL.
- Field-level permlevel filtering via `frappe.permissions.get_permitted_fields` (fails closed to permlevel-0 fields on any error — flagged for a bench smoke-test below).

## Frontend
- `FrappeListView` / `FrappeFormView` / `FrappeReportView` (`frontend/src/components/chat/frappe-views/`) — table/pagination via `@tanstack/react-table`, form fields (select/mandatory/read-only) driven by the meta payload.
- Wired into `ArtifactRenderer.tsx`'s existing type switch; new types added to `artifact.types.ts` and the `artifactParser.ts` allowlist.
- "Open in Desk" link on each view, using a doctype→slug helper (`frappeDeskUrl.ts`) since no existing one was found in-repo.
- `render_frappe_view` wired into the standard tool-discovery path (`_registry.py` → `huf_tools` hook → `sync_discovered_tools`) and described in `artifact_instructions.py` so an agent knows when to call it.

## Known follow-ups / deferred
- Report mode is currently `handle_list_records` against the doctype, not a real Query Report/Report Builder run — `erpnext_reports.py`'s report catalogue isn't disambiguated by type, so "Open in Desk" for report mode currently links to the doctype's list view rather than `/app/query-report/<name>`. Flagged with a TODO in `FrappeReportView.tsx`.
- `frappe.permissions.get_permitted_fields`'s exact signature on this codebase's Frappe version wasn't verified against a live install — recommend a bench-console smoke test before relying on it in production.
- No live bench/ERPNext end-to-end test was run in this environment (no bench available here); verification was code-level (permission-check tracing, `py_compile`, `tsc --noEmit` on touched files, manual cross-check of the artifact JSON contract between backend and frontend).
- Pagination re-fetch wiring (actually calling the tool again on page/filter change) is stubbed via `onPageChange`/`onFilterChange` props, not wired to a live round-trip yet.

## Track
Full dependency-mapped plan, phase-by-phase execution log, and the architecture review that shaped this (opus review of the original plan) are at `Tracks/safwan-erooth.DeskAIArtifactWorkspace/PLAN.md` in the workspace repo (not part of this diff).

### DIFF
--- a/frontend/src/components/chat/ArtifactRenderer.tsx
+++ b/frontend/src/components/chat/ArtifactRenderer.tsx
@@ -134,8 +134,8 @@ function renderFrappeView(type: 'frappe-list' | 'frappe-form' | 'frappe-report',
 	// while otherwise reproducing the payload faithfully). Recover the
 	// common single-key-rename case rather than showing a hard error for
 	// data that's actually all there.
-	if (payload && payload.data === undefined && (payload as Record<string, unknown>).doc !== undefined) {
-		payload = { ...payload, data: (payload as Record<string, unknown>).doc } as FrappeViewPayload;
+	if (payload && payload.data === undefined && (payload as unknown as Record<string, unknown>).doc !== undefined) {
+		payload = { ...payload, data: (payload as unknown as Record<string, unknown>).doc } as FrappeViewPayload;
 	}
 
 	if (!payload || !payload.meta || payload.data === undefined) {
--- a/frontend/src/components/chat/ArtifactRenderer.tsx
+++ b/frontend/src/components/chat/ArtifactRenderer.tsx
@@ -128,6 +128,16 @@ function renderFrappeView(type: 'frappe-list' | 'frappe-form' | 'frappe-report',
 		);
 	}
 
+	// Defensive normalization: a model asked to relay the tool's artifact tag
+	// "verbatim" can still occasionally paraphrase a key name (observed:
+	// gemini-3.6-flash emitting `doc` instead of `data` for mode="form"
+	// while otherwise reproducing the payload faithfully). Recover the
+	// common single-key-rename case rather than showing a hard error for
+	// data that's actually all there.
+	if (payload && payload.data === undefined && (payload as Record<string, unknown>).doc !== undefined) {
+		payload = { ...payload, data: (payload as Record<string, unknown>).doc } as FrappeViewPayload;
+	}
+
 	if (!payload || !payload.meta || payload.data === undefined) {
 		return (
 			<div className="text-sm text-destructive p-4">
--- a/frontend/src/components/chat/ArtifactRenderer.tsx
+++ b/frontend/src/components/chat/ArtifactRenderer.tsx
@@ -40,6 +40,11 @@ import { Mermaid } from '@/components/ui/mermaid';
 import { JSXPreview, JSXPreviewContent, JSXPreviewExport } from '@/components/ui/jsx-preview';
 import { Video } from '@/components/ai-elements/video';
 import { DocumentPreview } from '@/components/chat/DocumentPreview';
+import { FrappeListView } from '@/components/chat/frappe-views/FrappeListView';
+import { FrappeFormView } from '@/components/chat/frappe-views/FrappeFormView';
+import { FrappeReportView } from '@/components/chat/frappe-views/FrappeReportView';
+import type { FrappeViewPayload } from '@/types/artifact.types';
+import { TableIcon } from 'lucide-react';
 
 /** Ids minted by the client-side parser for transient, unsaved artifacts. */
 const TRANSIENT_ARTIFACT_ID = /^artifact-\d+-\d+$/;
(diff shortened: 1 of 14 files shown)


esafwan:
## Live end-to-end verification

Tested against a disposable bench (frappe15 + erpnext15 + huf branch `feat/frappe-artifact-views` + deskai), real ERPNext seed data, via the actual Chat UI (not just unit-level checks):

- **List view**: `render_frappe_view(doctype="Sales Invoice", mode="list")` → real customers (Acme Corp, Global Tech Ltd), sortable columns, pagination footer, "Open in Desk" link.
- **Form view**: `render_frappe_view(doctype="Customer", mode="form")` → full field-type-aware form (Select dropdowns with real options, Check checkboxes, Table-field placeholders, mandatory markers).
- **Report view**: `render_frappe_view(doctype="Item", mode="report")` → filter bar + table, "Open in Desk" (routed to the doctype list per the documented report/query-report ambiguity).
- **Chart artifact** (pre-existing feature): confirmed no regression — bar chart still renders correctly alongside the three new types.
- **Real pagination**: clicked "Next" on a 10-row Item list (page size 4) — advanced live from SKU001-004 to SKU005-008, backed by a genuine `frappe.client.get_list` refetch (not the static-snapshot stub the components originally shipped with).
- **Real filtering**: typed into the Report view's "Item Code" filter — debounced live re-query went from "No records" (unfiltered payload's default state) to 7 matching rows.

**Bugs found and fixed during this pass** (all pushed to this branch):
- `handle_render_frappe_view` was dropping `limit_start`/`limit_page_length` from the list/report envelope even though `handle_list_records` already computed them (`dffcf1ad`).
- A cheap test model (gemini-3.6-flash) asked to relay the artifact tag "verbatim" instead paraphrased the form payload's `data` key as `doc`, hitting a hard `meta/data` guard and showing an error card instead of the form. Added a defensive single-key normalization in `ArtifactRenderer` plus a stronger instruction against renaming keys (`208e1c03`, `539c2fda`).
- Pagination/filtering were originally client-side-only stubs (`onPageChange`/`onFilterChange` just reported the requested state with no actual refetch) — confirmed via a live click that did nothing. Rewired both `FrappeListView` and `FrappeReportView` to genuinely re-fetch via `frappe.client.get_list` (`db.getDocList`), the same whitelisted, permission-checked method already used elsewhere in the frontend (`162db723`).

Two pre-existing, unrelated bugs were hit and fixed *on the disposable test bench only* (not part of this PR's diff, since they're bugs in `pre-dev` itself): `resolve_model_context_window` returning `None` for an unlisted AI Model with no fallback before an `Agent Run` NOT NULL insert, and a schema-sync gap for `Agent Run.model_context_window`. Worth a separate fix upstream but out of scope here.

Screenshots attached to the user in-session; not re-uploaded here to avoid duplicating scratch-bench artifacts in the PR history.

esafwan:
Superseded by #650 — same commits, rebased/retargeted from pre-dev to pre-develop, plus DeskAI-side native-Desk rendering added.

PR #648, opened 2026-08-23
analytics: surface run duration in conversation analytics

esafwan:
## Follow-up to #645

Post-merge, a manual review of the conversation analytics response (added in #645) found that
duration was never surfaced despite `Agent Run.start_time`/`end_time` already existing and already
being aggregated by the scheduled rollup (`duration_ms_sum`/`duration_count`) — `get_conversation_analytics`
simply never fetched `end_time`.

- `totals.duration_ms_sum` / `duration_count` / `average_duration_ms` — cumulative, field names
  deliberately match `agent_run_analytics_api.py`'s summary shape rather than inventing a second
  convention for the same idea.
- `current.duration_ms` — the latest turn's own duration, a snapshot like everything else in
  `current`, never summed into `totals`.
- `series[].duration_ms` — per-run, for a future latency-over-turns view.
- Negative durations (clock skew / bad data) are treated as unmeasured, same guard the rollup already
  uses — never reported as a negative number.

Two new tiles in `ConversationAnalyticsPane`: "Avg duration" in Totals, "This turn" in Current
context, using the same `(ms/1000).toFixed(1)s` formatting `ExecutionAnalyticsDashboard.tsx` already
uses.

10 new standalone tests, all pass. Frontend typecheck clean. Live bench verification in progress —
will comment with results.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -57,7 +57,7 @@ const ChatScheduledPage = lazy(() =>
   import('./pages/chat/ChatPlaceholderPages').then((m) => ({ default: m.ChatScheduledPage }))
 );
 const Executions = lazy(() => import('./pages/Executions'));
-const AnalyticsPage = lazy(() => import('./pages/AnalyticsPage'));
+const AnalyticsEntityDetailPage = lazy(() => import('./pages/AnalyticsEntityDetailPage'));
 const AgentRunDetailPageWrapper = lazy(() => import('./pages/AgentRunDetailPageWrapper'));
 const McpDetailsPageWrapper = lazy(() => import('./pages/McpDetailsPageWrapper'));
 const McpListingPage = lazy(() => import('./pages/McpListingPage'));
@@ -539,13 +539,14 @@ function AppShell() {
               </ProtectedRoute>
             }
           />
+          <Route path="/analytics" element={<Navigate to="/executions?tab=analytics" replace />} />
           <Route
-            path="/analytics"
+            path="/analytics/:dimension/:entity"
             element={
               <ProtectedRoute>
                 <UnifiedLayout>
                   <Suspense fallback={<PageLoader />}>
-                    <AnalyticsPage />
+                    <AnalyticsEntityDetailPage />
                   </Suspense>
                 </UnifiedLayout>
               </ProtectedRoute>
--- a/frontend/src/components/app-sidebar.tsx
+++ b/frontend/src/components/app-sidebar.tsx
@@ -1,5 +1,5 @@
 import * as React from "react"
-import { ArrowLeft, Home, ChartColumnIncreasing, ChartNoAxesCombined, SquareAsterisk, FileText, Workflow, Database, Layers, MessageSquare, Zap, Server, Users, BookOpen, Link2, Terminal, Settings, LayoutGrid, Brain, Sparkles, SquareChevronRight, ChevronsLeftRightEllipsis, GlobeLock, Keyboard, SlidersHorizontal, type LucideIcon } from "lucide-react"
+import { ArrowLeft, Home, ChartColumnIncreasing, SquareAsterisk, FileText, Workflow, Database, Layers, MessageSquare, Zap, Server, Users, BookOpen, Link2, Terminal, Settings, LayoutGrid, Brain, Sparkles, SquareChevronRight, ChevronsLeftRightEllipsis, GlobeLock, Keyboard, SlidersHorizontal, type LucideIcon } from "lucide-react"
 import { useLocation } from "react-router-dom"
 
 import { NavMain } from "@/components/nav-main"
@@ -127,13 +127,6 @@ export const operateNavItems = [
 		title: "Executions",
 		url: "/executions",
 		icon: Zap,
-		capability: "agent.use",
-		badge: "Experimental",
-	},
-	{
-		title: "Analytics",
-		url: "/analytics",
-		icon: ChartNoAxesCombined,
 		capability: "agent.use",
 		badge: "Experimental",
 	},
--- a/frontend/src/components/chat/analytics/ConversationAnalyticsPane.tsx
+++ b/frontend/src/components/chat/analytics/ConversationAnalyticsPane.tsx
@@ -67,6 +67,10 @@ function formatPercent(ratio: number): string {
   return `${(ratio * 100).toLocaleString(undefined, { maximumFractionDigits: 1 })}%`;
 }
 
+function formatDurationMs(ms: number): string {
+  return `${(ms / 1000).toFixed(1)}s`;
+}
+
 function runKindBreakdown(byKind: Record<ConversationRunKind, number>): string {
   return (Object.keys(RUN_KIND_LABEL) as ConversationRunKind[])
     .map((kind) => [kind, byKind[kind] ?? 0] as const)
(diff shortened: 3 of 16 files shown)


esafwan:
## Live end-to-end verification against a real Google/Gemini provider

Synced this branch into an already-provisioned bench (`pre-dev-merged`, avoiding another full
`bench init` after a shared-host memory incident — see below), configured a real Google AI Provider
(key reused from another bench's DB, decrypted read-only) and `gemini-3.6-flash`, created a test agent,
and had an actual multi-turn chat conversation through the real browser UI.

**Real conversation, real Gemini responses:**
- "Hi! Briefly, what is 12 * 7?" → "12 * 7 is 84." (3,055 input / 81 output tokens, $0.002595, 1.65s)
- "Great. One more: what's the capital of France?" → "The capital of France is Paris." (3,080 input /
  34 output tokens, $0.0024375, 1.41s)

**The conversation analytics pane, opened live in the chat UI, rendered:**
- Totals: 2 runs, 6,135 billed input tokens, 115 output tokens, $0.005 cost, **Avg duration: 1.5s**
- Current context: 3,080 tokens largest turn, 0.3% of the 1,048,576-token window, **This turn: 1.4s**
- Composition breakdown, cache effectiveness (0% — no caching on these short single-round turns, correctly
  labeled as an indicator), "Every run in this conversation was fully measured"
- The context-growth chart rendered with real per-turn data points

**With vs without, on the exact same real data:** ran these same two real `Agent Run` rows through both
the pre-fix code (`git show 512be706:.../agent_conversation_analytics_api.py`, the #645 merge commit)
and this branch's code, in the same bench console:

```
BEFORE (#645 as merged): any 'duration' field present? False
AFTER  (this branch):    duration_ms_sum=3060.02  duration_count=2  average_duration_ms=1530.01
                          current.duration_ms (latest turn) = 1410.194

Cross-check against the live browser pane:
UI showed 'Avg duration: 1.5s'  vs computed 1.5s
UI showed 'This turn: 1.4s'     vs computed 1.4s
```

Exact match between the standalone computation and what actually rendered in the browser.

**Aside, disclosed for transparency:** provisioning a fresh bench for this test triggered an OOM crash
of the whole shared devcontainer host (too many concurrent bench builds for its ~15.7GB cap), which
briefly took production down too (~1 minute, clean DB shutdown, auto-recovered, verified healthy).
Switched to syncing an already-built bench instead of building a new one, which resolved it. Two
incidental fixes made in the process, unrelated to this PR's diff: enabled `allow_chat` on the AI-answer
gate for the AI Model default seed, and registered the site hostname in the container's `/etc/hosts`
(a documented gotcha, `frappe-multihand` skill §3.5) — neither is a code change, both were bench-local
config.

esafwan:
Small correction to the comment above: the `allow_chat` flag I enabled was on the test **Agent** (`manual-test-agent`), not on the AI Model — that's what controls whether an agent shows up in the chat picker. Both incidental fixes were bench-local config, not code changes.

esafwan:
## Two more real bugs found live-testing this branch, both fixed

Testing \`/analytics\` in the browser (not just via curl/console) surfaced two crashes that only show
up with real, non-empty data — exactly what the earlier live verification's empty test bench never
exercised.

### 1. \`get_execution_analytics\` 500s: naive/aware datetime compare

\`TypeError: can't compare offset-naive and offset-aware datetimes\`, on every single \`/analytics\` page
load. \`AnalyticsPage.tsx\` always sends \`from_date\` via \`Date.toISOString()\` (a \`Z\`-suffixed string,
which \`get_datetime()\` parses as timezone-**aware**) and never sends \`to_date\` (which falls back to
\`now_datetime()\`, always **naive**). This bug predates this branch — confirmed present on
\`origin/pre-dev\` before #645 touched this file — but #645 is what shipped the page that triggers it
unconditionally.

The fix isn't "just strip tzinfo" — this site's configured timezone is Asia/Kolkata (UTC+5:30), not
UTC, so a blind strip would silently shift the query window by 5.5 hours instead of crashing loudly.
Routed through Frappe's own \`convert_utc_to_system_timezone\` (reads System Settings, not a hardcoded
zone, so it's correct for any deployment/timezone, not just this one). Verified: a request with
\`from_date=2026-08-17T06:58:19.066Z\` now correctly resolves to \`2026-08-17 12:28:19\` local
(+5:30 exactly).

### 2. Every breakdown/series row crashed the page once real data existed

\`Cannot read properties of undefined (reading 'toFixed')\` — the instant the rollup table had any real
rows, the \`/analytics\` page's breakdown table crashed entirely. \`success_rate\`/\`average_duration_ms\`/
\`cache_ratio\` were only ever computed on the top-level \`summary\` dict, **after** the accumulation loop
that seeds \`series_by_bucket\`/\`breakdown_by_dimension\` from summary's *then-current* keys — so those
three keys were entirely **absent** (\`undefined\` in JS, not \`null\`) on every series bucket and
breakdown row. The frontend's own type (\`ExecutionAnalyticsResponse\`) already declared the real
contract: \`series\`/\`breakdowns\` are typed as \`Array<ExecutionAnalyticsSummary & {...}>\`, the *same*
type as summary, derived fields included — the backend just wasn't upholding its own declared shape.
\`row.success_rate === null\` doesn't catch \`undefined\`, so it fell through to \`undefined.toFixed(1)\`.

Extracted \`_add_derived_rates()\`, now called on \`summary\` **and** every series bucket **and** every
breakdown row.

Both fixes verified live: seeded a real bench with an actual Google/Gemini conversation (4 real runs),
confirmed the exact same crash reproduces pre-fix, and confirmed the full \`/analytics\` page — summary
tiles, breakdown table, context-composition chart — renders correctly post-fix, including switching
time windows (a genuinely fresh request, not a cached one).

3 new regression tests for \`_add_derived_rates\` (always-present keys even at zero denominators, real
ratio computation, breakdown-row shape). 91/91 standalone tests pass; frontend typecheck clean.

esafwan:
Correction: the actual count is **93/93** standalone tests passing, not 91 — miscounted in the previous comment.

esafwan:
## Phase 7 — Entity drill-down, bidirectional linking, IA consolidation

Adds the 4 follow-on items requested after live-testing the analytics page:

1. **Entity drill-down** — clicking an agent/provider/model row in the Analytics breakdown table opens `/analytics/:dimension/:entity`, a new page (`AnalyticsEntityDetailPage.tsx`) scoped to just that entity: summary tiles, a runs-over-time trend chart, and the same `ContextBar` composition visualization already used on run detail pages (reused as-is, no new viz code). Backed by a new optional `entity` param on `get_execution_analytics`.
2. **Bidirectional conversation linking** — clicking a conversation row instead navigates to `/chat/:id?pane=analytics`, which auto-opens the existing per-conversation analytics pane (no new page — the pane already existed).
3. **IA consolidation** — `/executions` and `/analytics` are now tabs ("Runs" / "Analytics") of one page, tab state synced to `?tab=`, single sidebar entry; `/analytics` becomes a client-side redirect so old links keep working.

### How it was built
Executed as 4 dependency-mapped waves (9 tasks) via parallel haiku/sonnet-5-low subagents, diff-checked between waves, tracked in `Tracks/safwan-erooth.AnalyticsObservability/TRACKER.md` (Phase 7 section). A final adversarial sonnet-5 review pass caught and fixed one real bug before this push: `AnalyticsEntityDetailPage.tsx` was double-decoding the `entity` route param (React Router already decodes it once), which would throw/corrupt on any entity name containing a raw `%`.

### Verified live
Synced into the `pre-dev-merged` bench, rebuilt, migrated, restarted clean. Manually clicked through in-browser:
- Analytics tab renders inside the merged Executions page, provider breakdown → clicking "Google" row navigates to `/analytics/provider/Google` and renders tiles + trend chart + context composition correctly.
- Switched dimension to "Conversation", clicked a conversation row → navigated to `/chat/<id>?pane=analytics`, chat loaded with the analytics pane auto-opened (Totals/Current-context tiles including the earlier avg-duration fix).
- `yarn typecheck` clean; new backend tests for the `entity` filter (6 cases) plus existing derived-rates regression tests all pass.

Commits: `445c34b2` (Phase 7 implementation), `755f7345` (double-decode fix from final review).

PR #650, opened 2026-08-24
feat: render ERPNext/Frappe list/form/report views inline as chat artifacts + native Desk panels via DeskAI

esafwan:
Supersedes #647 (retargeted from `pre-dev` to `pre-develop` per updated instructions — same commits, clean rebase, no `pre-dev`-only history mixed in).

## Summary
Two integration surfaces for the same capability — rendering generic Frappe/ERPNext data (list, form, report) without the user leaving chat, or inline inside Desk itself:

1. **huf chat** — new artifact types (`frappe-list`/`frappe-form`/`frappe-report`), rendered via `ArtifactRenderer.tsx`'s existing `ai-elements` `Artifact`/`ArtifactHeader`/`ArtifactContent` wrapper (same primitive every other artifact type uses — no new panel component).
2. **DeskAI** (separate Frappe app, `esafwan/deskai`, runs inside authenticated Desk) — new `render_frappe_list`/`render_frappe_report` client tools that render inline panels directly inside Desk using Frappe's own DOM/CSS classes, so styling is native Desk, not a port.

Both are **generic across any Frappe app's doctypes** — not ERPNext-specific. Verified against ERPNext doctypes (Sales Invoice, Item) only because that's what's installed on the test bench; the tool takes an arbitrary `doctype`/`report_name` string.

No `frappe-ui-react` or `frappe-ui` dependency was added or ported from — evaluated and rejected per an opus-reviewed architecture pass (community port, second styling/data stack). huf's components use the existing design system + `@tanstack/react-table`; DeskAI's panels reuse Desk's own `table`/`btn`/`form-control` classes directly.

## Backend (`huf/ai/tools/frappe_generic.py`, new)
- `handle_get_doctype_meta`, `handle_list_records`, `handle_get_record` — generic reads via `frappe.get_meta`/`frappe.get_list`/`frappe.get_doc`.
- `handle_create_record`/`handle_update_record` — **draft-only**, never write server-side; return a draft payload, the real write happens client-side as the logged-in user through the normal Frappe REST path.
- `handle_render_frappe_view` — composes the above into the three new artifact types.
- `erpnext_reports.handle_run_report` extended with `limit_start`/`limit_page_length` pagination.

### Security (non-negotiable, enforced in every handler)
- Doctype denylist: `User`, `Role`, `Role Profile`, `Custom Script`, `Server Script`, `Property Setter`, `File`, any `OAuth*`/`Integration*` prefix, all Singles — deny-by-default if a doctype doesn't exist or its meta can't load.
- `frappe.has_permission(doctype, ptype=...)` enforced per operation; reads go through `frappe.get_list`/`frappe.get_doc` (permission-query conditions applied), never raw SQL.
- Field-level permlevel filtering via `frappe.permissions.get_permitted_fields` (fails closed to permlevel-0 fields on error).
- DeskAI's client tools carry the same doctype denylist and rely on Frappe core's own permission enforcement on `frappe.client.get_list`/`frappe.desk.query_report.run` (DeskAI runs as the real logged-in Desk user, so no separate permission layer needed there).

## Frontend (huf)
- `FrappeListView`/`FrappeFormView`/`FrappeReportView` (`frontend/src/components/chat/frappe-views/`) — `@tanstack/react-table`-based, meta-driven field rendering (select/mandatory/read-only), real pagination/filter refetch (not a stub — this was caught and fixed once during testing, see commit `ddf5af65`).
- "Open in Desk" deep link on each view — same-site deployment confirmed (huf and Desk share a site/session), so a plain `/app/...` link just works, no SSO handoff needed.
- `render_frappe_view` wired into the standard tool-discovery path and described in `artifact_instructions.py`.

## DeskAI (separate repo, `esafwan/deskai`, commit `01a5b82`)
- `render_frappe_list`/`render_frappe_report` added to the widget's `DeskTools` registry, registered with the backend agent in `setup_assistant.py`.
- Inline panels built with plain DOM (`makeDataPanel`/`buildTable`), real refetch on pagination (`refetchList` → `frappe.client.get_list`) and filter (`refetchReport` → `frappe.desk.query_report.run`) — confirmed live network round-trips, not UI-only stubs.
- Same doctype denylist as the huf backend.

## Testing
Live-tested on a disposable frappe15+erpnext15+huf+deskai bench (`deskai-erpnext`):
- huf chat: frappe-list, frappe-form (including a bug caught and fixed where a model paraphrased `data` as `doc` — commits `f3a99f0f`/`15ff8433`), frappe-report, and existing chart artifacts — screenshotted.
- Real pagination click (Next) on a list view — verified row set actually changed, not a static re-render.
- Real filter interaction on a report view — verified live re-query (row count changed).
- DeskAI-side native-Desk rendering — in progress at PR open time; screenshots to follow in a review comment once that pass completes.

## Known follow-ups
- Report mode's "Open in Desk" link goes to the doctype's list view, not `/app/query-report/<name>` — `erpnext_reports.py`'s report catalogue doesn't disambiguate Query Report vs. Report Builder, so this wasn't guessed at.
- `frappe.permissions.get_permitted_fields`'s exact signature wasn't verified against a live bench beyond the smoke test already run — recommend an extra pass before wide rollout.
- No visual token pass was done to pull frappe-ui's actual design tokens (spacing scale, color variables) into huf's components — huf's own design system was used as-is for consistency with the rest of huf's UI, and DeskAI's panels inherit Desk's CSS natively. Flagged as an explicit follow-up, not silently skipped.

Full dependency-mapped plan and phase-by-phase execution log: `Tracks/safwan-erooth.DeskAIArtifactWorkspace/PLAN.md` (workspace repo, not part of this diff).


esafwan:
## Live E2E test results (disposable frappe15+erpnext15+huf+deskai bench)

Tested both integration surfaces this feature ships: **huf chat** and **DeskAI running natively inside Frappe Desk**.

### huf chat (Demo Assistant agent, `render_frappe_view` tool)
- `frappe-report` (Item, with filters bar, "Open in Desk" link, 10 records, paginated) — confirmed working end to end.
- `frappe-list`, `frappe-form` (proper select/mandatory/read-only field rendering, child-table row counts), and a `chart` artifact (pre-existing type, confirmed not regressed) — all confirmed rendering.
- Applied a styling pass (commit `f7283057`) after the user asked for Desk/frappe-ui visual consistency: column headers and filter labels now use Desk's small uppercase-mono-gray convention, row density and control sizing tightened to match Desk instead of reading as a generic huf card. Composed entirely from huf's existing design tokens — no new dependency, no `frappe-ui`/`frappe-ui-react` import. `tsc --noEmit` clean. Re-screenshotted to confirm the visual change took effect.

### DeskAI inside real Desk (`/app`, floating widget, separate `esafwan/deskai` repo)
This is the "render inline where the user already is" half of the original ask — not just huf's own chat UI.
- `render_frappe_list` against "Item" from the Desk Home page: real inline table, "Open in Desk" link, then a genuine Next-page click (SKU001-005 → SKU006-010) — confirmed to be a real refetch, not a stub.
- `render_frappe_report` against an invalid report name correctly returned `ok:false` (proper error handling, not a silent failure); the underlying `frappe.desk.query_report.run` call was separately confirmed working via a direct console call. A live success-case screenshot against a real report name wasn't captured — the small test model (gemini-3.5-flash-lite) looped on tool calls rather than committing to a report name, which is a model-choice issue in this bench's config, not a defect in the tool.

**Two real bugs found and fixed during this pass** (both in the DeskAI repo, not huf):
1. `deskai.api.todo_write_broadcast` was missing `@frappe.whitelist()` — its own doctype validation rejected it as unwhitelisted, which aborted the *entire* `bench migrate` after_migrate seeding transaction with a non-zero exit before anything (including our new tools) could commit. Fixed.
2. `checkDoctypeRenderAllowed` called `frappe.get_meta(doctype)` synchronously, which only has data for doctypes Desk has already lazily loaded client-side — asking for "Items" from the Home page (never visited) failed with `"meta not loaded for DocType: Item"`. Fixed by awaiting `frappe.model.with_doctype` first, with a 10s timeout fallback so an unknown/inaccessible doctype can't hang the tool call forever.

### Screenshots
File-based screenshots (huf chat side) are in the track directory `Tracks/safwan-erooth.DeskAIArtifactWorkspace/screenshots/` (not part of this diff — workspace-repo only): report view (styled + pre-style), form view, list view, chart, pagination. DeskAI-inside-Desk proof is inline in this session's transcript — no file-capture mechanism was available for that interactive pane in this environment.

### Not yet covered
- Query Report success-case screenshot from DeskAI (see above).
- Live bench test of the `handle_create_record`/`handle_update_record` draft-then-user-confirms write path (code-reviewed, not click-tested).
- Cross-browser/dark-theme check of the new mono-uppercase label styling.

PR #656, opened 2026-08-24
Automation triggers: audit fixes, send_email tool, chat entry point, two live-verified trigger bugs

esafwan:
## Summary

Audit + fixes for the DocType-event → Agent Automation trigger system, prompted
by: "when a Lead is created with a contact email, the agent should automatically
send an acknowledgement email" not being achievable, plus a UI bug report
(automation form page not scrollable, save silently failing).

Live-verified end-to-end against a disposable bench (real HTTP requests + RQ
worker, not just static review) — two real bugs were found and fixed as part of
this.

## What's in this PR

- **Dead code removal**: unreferenced legacy `TriggersTab.tsx`/`TriggerModal.tsx`
  (superseded by the `Automation` doctype UI), -714 lines.
- **`Send Email` built-in Agent Tool Function**: reuses `frappe.sendmail`,
  validates recipient addresses, checks read permission on a threaded reference
  doc. Optionally routes through a configured **Email Gateway**'s sender address
  (`gateway` param) instead of the site default outgoing account, reusing the
  existing `get_gateway_adapter()` helper — so an Automation can send from a
  specific connected inbox.
- **"Create automation from this chat"**: explicit chat-header menu action that
  pre-selects the conversation's Agent on the existing automation form. No new
  LLM-calling pathway invented — deliberately scoped as a mechanical pre-fill,
  not automated drafting.
- **Runtime-mode visibility**: a legacy-only warning banner on the Automations
  tab, so a site accidentally left on the legacy trigger runtime doesn't fail
  silently with no explanation.
- **Automation form scroll + tabs fix**: the page's root container was missing
  the `h-full overflow-auto` wrapper every sibling form page has, so the
  required "Instruction" field (and the whole Trigger/Execution sections) were
  unreachable below the fold — this is why Create silently "did nothing" (it
  was failing validation on a field the user could never scroll to). Also
  grouped General/Task/Trigger/Execution/History into tabs per request.
- **Bug fix — Doc Event triggers could silently never fire**: `hooks.py`'s
  cache-invalidation entry for the trigger cache was wired for the legacy
  `Agent Trigger` doctype only, never for the new `Automation Trigger` doctype.
  Once the cache got populated empty for an event (e.g. by a scheduler tick
  before any trigger existed), every Automation Trigger created afterward for
  that event was invisible forever, with no error anywhere. Reproduced and
  confirmed fixed live.
- **Bug fix — Schedule triggers never got a first run time**: a new
  Schedule-type `Automation Trigger` never had `next_execution` computed, and
  the scheduler's `next_execution <= now` filter excludes SQL NULL, so it sat
  forever without ever becoming due. Now computed on `validate()` when unset.
  Confirmed fixed live.
- **Unrelated one-line fix**: removed an unused `ChartNoAxesCombined` import in
  `app-sidebar.tsx` that was already broken on `pre-develop`'s tip and blocking
  the real `yarn build` (not just `tsc --noEmit`) for any PR built on top of it.

## Test plan

- [x] `yarn typecheck` / `tsc -b` real build clean
- [x] Provisioned a disposable bench off this branch, live-created a Doc Event
      Automation (Lead/ToDo → agent) and a Schedule Automation via bench
      console + real `frappe.client.insert` HTTP calls (not just console, since
      console sessions don't reliably drive `after_commit` callbacks)
- [x] Reproduced the stale-cache bug before the fix, confirmed the fix resolves it
- [x] Confirmed a new Schedule trigger now gets a real `next_execution`
- [x] Confirmed the full pipeline fires through to `run_agent_sync`, failing only
      on "Agent model is not configured" (expected — disposable bench has no
      real AI Provider credentials, not an app bug)
- [ ] Not tested: actual email delivery (no SMTP configured in the test bench),
      Webhook/App Event trigger types, real LLM execution with live credentials

## Known follow-ups (not in this PR)

- `send_email` not wired into Flow Engine or Skills tool registries (scoped to
  Automations/Agent only, per the original request)
- "Duplicate automation" UI action is a pre-existing stub (`// TODO S17`)
- No admin-side pinning mechanism for the `gateway` param yet (unlike
  `reference_doctype`, which Agent Tool Function already supports pinning) —
  today it's just another LLM-suppliable/admin-defined tool parameter


esafwan:
## Round 3: deep live-testing results (email, webhook, app event, real LLM, frontend UI)

Per follow-up request, tested every remaining gap against the live `automation-e2e` bench.

**Email delivery** — stood up MailHog, confirmed real delivery via `send_email` (direct + via the async trigger path). Also added: `send_email` now accepts an optional `gateway` param to route the sender address through a configured Email Gateway (`e12ea6a9`), reusing the existing `get_gateway_adapter()` helper.

**Webhook trigger** — confirmed working end-to-end via a real unauthenticated POST with `X-Webhook-Key`. Needs `run_as_user` set on the Automation for a Guest caller to actually resolve the Agent — this is correct/designed behavior (confirmed by reading `automation_runner.py`), not a bug.

**App Event trigger** — the dispatcher (`trigger_app_event`) works correctly (confirmed via direct call and authenticated HTTP), but **is never called from any gateway adapter or integration** — no WhatsApp/Slack/Telegram inbound event fires it today. Real feature gap, not a bug (nothing's broken, it's just unbuilt wiring) — flagged for a future track, not fixed here.

**Live LLM** — real Gemini 3.6 Flash key configured, full loop confirmed: ToDo created via real HTTP insert → Doc Event trigger fires → Gemini call → `update_todo_status` tool call → ToDo flips to Closed, ~6s end-to-end.

**Frontend UI** — drove the real `/automations/new` page via Playwright: scroll fix confirmed (all tabs fit without scrolling), trigger-type dropdown confirmed showing all 5 types, DocType/event pickers confirmed working, automation created successfully via the UI.

**New findings (not fixed in this PR, flagged for follow-up):**
1. **Doc-event allowlist mismatch**: the frontend's Doc Event dropdown offers `after_submit`, but the backend's `automation_hooks.SUPPORTED_DOC_EVENTS` doesn't include it — a trigger configured for it would silently never fire. Same failure class as the cache-invalidation fix in this PR, different code path.
2. **Draft-status automations still fire**: a newly created Automation defaults to `status=Draft`, but Doc Event triggers fire identically regardless of status — `status` is purely informational. Not a functional bug, but surprising; worth a UI hint.

Full detail in `Tracks/safwan-erooth.AutomationAgentTriggers/TRACKER.md` (internal tracker, not part of this PR).

esafwan:
## Round 4: both flagged gaps fixed and live-verified

**Fix 1 — \`after_submit\` mismatch**: removed from the frontend's Doc Event dropdown (\`TriggerFieldsConfig.tsx\`). It was never in the backend's \`automation_hooks.SUPPORTED_DOC_EVENTS\`, so a trigger configured for it would save successfully and then silently never fire.

**Fix 2 — status was decorative, Draft/Paused/Archived automations fired anyway**: turned out to be bigger than just "Draft" — investigating found that \`pause_automation\`/\`archive_automation\` only ever set the \`status\` field, never the separate \`disabled\` flag that \`run_automation()\` actually checks. So **Pause literally did not pause anything** — a Paused or Archived automation's triggers kept firing exactly like an Active one.

Fixed in \`run_automation()\` (\`654b5795\`): automatic triggers (Doc Event/Schedule/Webhook/App Event — anything supplying a \`trigger_name\`) now require \`status == "Active"\`. Manual "Run now" and chat-triggered runs are exempt (\`trigger_name=None\`), so a Draft automation can still be tested before activation via the existing Resume button (confirmed the UI already has this — \`resume_automation\`'s own docstring says "Resume a paused (or draft/error) Automation", so this was clearly the intended design, just never wired to actually block execution).

**Live-verified**: activated one automation to Active (via \`resume_automation\`) and left a second one on Draft, then fired a real ToDo insert. Active automation closed the ToDo in ~7s as before. Draft automation was correctly blocked — \`total_runs\`/\`last_execution\` unchanged, and a new Error Log entry confirms the exact new message: *"Automation 'Close new Todos (UI test)' is not Active (status: Draft) -- resume it to let triggers run."*

esafwan:
## Round 5: point-4 file-attachment audit (OCR/audio) + standalone Automations page

**File attachments on Doc Event triggers — found and fixed a real bug** (`8dfb28cd`): `automation_hooks.py`'s Doc Event file-attachment path (OCR + audio transcription) hardcoded `agent_name=None` at both call sites into `handle_ocr_document()`/`transcribe_audio_file()`, both of which require a real agent_name to resolve provider/model config and immediately return `success=False` without it. So OCR/transcription **always failed silently, in every environment**, regardless of OCR/STT provider config — not a missing-dependency issue. Worse: the failure had zero trace anywhere (no Error Log, no exception — `if result.get("success")` just silently skipped it).

Fixed: resolve the Automation's linked Agent once, pass it through both calls. Added warning-level logging on a `success=False` result so this failure mode is diagnosable going forward. **Live-verified with a real file upload + Doc Event trigger** (not console-only): before the fix, the agent's response showed `"text": null` — after, it correctly quoted the exact OCR-extracted file content.

**Standalone `/automations` page** — you flagged that Automations was only reachable via an Agent's own tab, unlike Skills/Intelligence/Knowledge/Prompts which all have a top-level nav entry. Added:
- `AutomationsPage.tsx` — lists every automation across all agents (search, status filter, run now / pause-resume / archive), mirroring `SkillsPage.tsx`'s conventions.
- Sidebar entry under Build, next to Agents/Flows/Skills.
- Extracted the shared status-badge/timestamp/trigger-label helpers out of `AutomationsTab.tsx` into `utils/automationDisplay.ts` so the tab and the new page don't duplicate them.
- The automation detail page's breadcrumb now links back to `/automations`.

Live-verified in browser: page loads, shows all 5 test automations with correct status/trigger/agent/last-run, status filter dropdown works, sidebar entry present and navigable.

Also live-verified two earlier follow-up items from Round 3:
- **Real cron-driven Schedule trigger**: enabled the bench scheduler (was off by default), armed a trigger ~90s out, confirmed it fired via the actual cron tick (not a manual call) — `next_execution` correctly advanced by the configured interval.
- **3 trigger types on one Automation** (Doc Event + Webhook + App Event): all three fired correctly via their real transports. Hit one narrow, low-confidence timing race where a Doc Event trigger's cache lookup missed a just-created trigger under rapid back-to-back creation — force-clearing the cache and retrying fixed it immediately. Noted as a low-reproducibility observation, not chased further.

PR #661, opened 2026-08-24
fix(acp-federation): scaffolding — SSRF/perm fixes, no crash 

esafwan:
Backend scaffolding for a future "one Huf agent calls an agent on another Huf instance over HTTP" feature. **This is not a complete feature and should not be read as one** — see "what's still missing" below. What it does provide, safely: the DocType, the SSRF-hardened HTTP client, and the discovery/connection-test flow.

## What changed since the last review round

A second, deeper review pass found this PR was less complete than its name suggested — most of the RFC it's based on (`docs/rfcs/huf-remote-agent-federation.md`) was never finished. Rather than try to complete the whole RFC (real net-new feature work — 3 more DocTypes, a full run lifecycle, a frontend page), this round made what's *actually here* honest and non-crashing:

1. **Closed the SSRF bypass** — `test_connection()`/`refresh_manifest()` on the DocType previously called `requests.get()` directly, skipping all the SSRF hardening. They now go through the same validated `HufNativeAdapter._request()` path everything else uses.
2. **Manifest no longer lies about streaming** — it advertised `"streaming": true`; the RFC itself says V1 has none. Fixed to `false`.
3. **Added the missing permission checks** on `list_agents`/`get_agent_manifest` — every other whitelisted method in this PR already checked permissions, these two didn't.
4. **Fixed a read-permission-triggers-a-write bug** — a read-only role could hit `test_connection_cmd` and have it silently persist health-status fields via `ignore_permissions=True`. Now requires write permission, matching what the action actually does.
5. **Replaced a crash with an honest error** — the one code path that would actually execute a remote tool call (`tool_handler.py`) references a `Delegated Agent Run` doctype that was never built (it's one of the 3 missing RFC doctypes). It would have thrown an opaque `DoctypeNotFoundError`. It now checks first and raises a clear `RemoteAgentNotImplementedError` explaining remote tool-call execution isn't available yet.
6. **Fixed a silently-broken config option** — the DocType accepted `ws://`/`wss://` URLs, but the HTTP-only adapter would always reject them at request time with no clue why. Now rejected at save time with a clear message.
7. **Labeled the fake standards support clearly** — `AgentCommunicationProtocolAdapter` and `AgentClientProtocolAdapter` (meant for future real-ACP/A2A-standard interop) are pure `NotImplementedError` stubs. They now say so in their docstrings, so nobody mistakes this for real ACP interop.

## What's still missing (confirmed via interop testing, see below)

- **3 of 4 RFC doctypes were never built**: `Delegated Agent Run`, `Remote Agent Capability`, `Remote Agent Policy`. Without them, remote tool-calling from an agent's tool-call loop doesn't work — `tool_handler.py` now fails cleanly instead of crashing, but it still doesn't *do* anything.
- **`huf/api/remote_agents.py`'s `get_run`/`get_run_events`/`cancel_run`/`create_run` are stub endpoints** that don't touch any real run record — untouched in this pass, since fixing them needs the missing doctypes too.
- **No frontend UI exists** — configuration is desk-form-only.
- **This is not real ACP.** Tested directly: the actual Agent Client Protocol (used by e.g. `kimi-code acp`, Zed's agent integration) is JSON-RPC 2.0 over stdio with a rich capabilities handshake. What's implemented here (`HufNativeAdapter`) is a Huf-proprietary HTTP+manifest protocol that only borrows the "ACP" name — confirmed by directly running `kimi acp` and diffing its wire protocol against `adapter.py`'s.
- **What DOES work, verified**: wrote a standalone functional test (`test_huf_federation_interop.py`) against a mock second-Huf-instance HTTP server, exercising `HufNativeAdapter` directly (it has no Frappe dependency, so this is testable outside a bench). All 6 checks passed: manifest discovery, Frappe-RPC-envelope unwrapping, a full remote-agent-invocation round trip, the SSRF guard correctly rejecting loopback when `allow_local_network=False`, and bearer-token auth header construction. **The transport/adapter layer is solid** — what's broken is everything above it (missing doctypes, stub endpoints, no UI).

## Recommendation

Land this as what it honestly is: a safe, tested HTTP client + connection-management DocType for a federation feature that isn't finished. Rename/re-scope expectations before merge (title/description already reflect this). Follow-up work needed before "remote agent tool-calling" is real: the 3 missing doctypes, real `remote_agents.py` endpoints, `Agent Tool Function`'s "Remote Agent" tool_type wiring, and a frontend config page.

*Second-round fixes applied by an automated audit/fix pass over 2026-08-25's revived-branch inventory.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,7 @@ Upstream incident thread: https://github.com/BerriAI/litellm/issues/24518
 
 - [**Full Documentation**](https://docs.huf.ai/) — Guides, tutorials, and API reference
 - [**Bruno API Collection**](./docs/bruno/) — Ready-to-use API collection for testing and exploration
+- [**Remote Agent Federation RFC**](./docs/rfcs/huf-remote-agent-federation.md) — Protocol, schema, API & runtime specification for HUF remote agent federation.
 - [**AGENTS.md**](./AGENTS.md) — Technical context for AI agents. Adopts the [agents.md](https://agents.md) standard.
 - [**CLAUDE.md**](./CLAUDE.md) — Defines coding standards, review criteria, and project-specific rules.
 
--- a/docs/rfcs/huf-remote-agent-federation.md
+++ b/docs/rfcs/huf-remote-agent-federation.md
@@ -1,0 +1,430 @@
+# RFC: HUF Remote Agent Federation Protocol & Runtime Specification
+
+- **Status**: Proposed / Draft (Phase 1 Specification)
+- **Author**: HUF Architecture Team
+- **Created**: 2026-08-01
+- **Target Version**: V1.0
+- **RFC File**: `docs/rfcs/huf-remote-agent-federation.md`
+
+---
+
+## 1. Executive Summary
+
+HUF serves as the Frappe-native AI control plane, orchestrating AI agents, tool calls, workflow triggers, scoped memory, and enterprise data models within Frappe. To support interoperability with external agent ecosystems (e.g., ACP-compliant external tools), editor/IDE coding agents (e.g., Claude Code, Cursor, OpenCode), and satellite HUF nodes without altering or replacing HUF's native internal state machine, HUF introduces an **Adapter-based Remote Agent Federation** model.
+
+Under this architecture:
+> **HUF remains the Frappe-native AI control plane.** Protocols such as Agent Communication Protocol (ACP), Agent Client Protocol, and HUF-Native Federation act as adapter boundaries for calling remote agents, exposing local HUF agents, and linking satellite HUF instances.
+
+This specification documents Phase 1 of the federation roadmap, detailing protocol positioning, V1 scope boundaries, new Frappe DocTypes, REST APIs, the normalized event lifecycle, security governance rules, and a two-node demo rollout plan.
+
+---
+
+## 2. Protocol Positioning & Framing
+
+There are distinct protocol standards in the modern AI ecosystem. HUF positions itself cleanly across each boundary without conflating their roles:
+
+| Protocol | Purpose & Target Domain | HUF Architectural Role |
+| :--- | :--- | :--- |
+| **Model Context Protocol (MCP)** | Tool, resource, and prompt access between host and server. | HUF acts as an MCP client today (consuming external MCP tools). MCP server/gateway functionality remains a separate service boundary. |
+| **Agent Communication Protocol (ACP)** | High-level agent-to-agent interoperability via RESTful APIs, manifests, async runs, streaming, and sessions. | Supported via an adapter for delegating tasks to external ACP-compliant agents and exposing select HUF agents to external orchestrators. |
+| **Agent Client Protocol** | IDE / Editor to coding agent protocol (e.g., Claude Code, Cursor, OpenCode, Codex). | Supported via an adapter for delegating code manipulation tasks, terminal command execution, and workspace diff reviews. |
+| **HUF-Native Federation Protocol** | High-efficiency node-to-node protocol between HUF instances. | Native Frappe REST protocol preserving full Frappe session context, granular permission policies, and audit correlation IDs across nodes. |
+
+```text
+               +----------------------------------------+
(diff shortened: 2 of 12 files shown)


esafwan:
## Automated review — needs changes before merge

Revived-branch PRs get zero human eyes by default, so this got a security-focused review pass before anyone looks at it.

### Blockers
- **`remote_agent_connection.py` `test_connection()`/`refresh_manifest()` bypass all SSRF hardening.** They call `requests.get(f"{base_url}/.well-known/huf-agent.json", ...)` directly instead of going through `adapter.py`'s `validate_remote_url`/`HufNativeAdapter`. A System Manager hitting "Test Connection" against `http://169.254.169.254/...` (or any internal service) gets the response back in the UI — full SSRF via a button in the admin UI. Route both through the same validated request path `_request` uses.
- **`remote_agents.py` `create_run()` is a stub that always returns fake success** (`run_id: "dummy_run_123"`) regardless of input. A remote instance calling into this feature gets a convincing but entirely fabricated "it worked." Needs to actually dispatch a run or fail loudly, not silently lie.

### Major
- **DNS-rebinding TOCTOU in `validate_remote_url`**: it resolves + validates a hostname, then `_request` lets `requests` re-resolve the same hostname independently at connect time. An attacker controlling DNS can pass validation with a public IP and connect to a private one moments later. Needs to validate and connect to a pinned IP, not re-trust a hostname string.
- **`list_agents`/`get_agent_manifest` have no permission check** beyond `allow_guest=False` — any authenticated user gets the full agent list + capabilities. Contrast with `tool_handler.py` and `remote_agent_connection.py` in this same PR, which correctly call `frappe.has_permission`/`doc.check_permission()`. Add the same check here.

### Minor
- `adapter.py`'s `_get_headers` supports `auth_type == "api_key"` but the DocType's Select field only offers `none/bearer_token/site_token` — dead branch, either wire it up or drop it.
- Trailing whitespace in `tool_handler.py` (would likely fail a linter).

### Good
- Credential storage follows house convention: `auth_secret` is a Frappe `Password` field, `as_dict()` strips it defensively, only accessed via `get_password`.
- DocType permission tiering (System Manager/Huf Manager full CRUD, Huf User read-only) looks sensible; nothing here should break `bench migrate`.
- Exception hierarchy in `adapter.py` is clean; auth-failure and timeout paths are well-tested.
- Redirect-based SSRF is blocked (`allow_redirects=False`).

**Test gap to flag**: every test class except the pure `validate_remote_url` unit tests mocks the validator to always pass — so none of the above SSRF-bypass surface has any chance of being caught by CI as currently written.

**Verdict: needs a deeper security review before merge**, specifically closing the `test_connection`/`refresh_manifest` bypass and the DNS-rebinding gap.

esafwan:
## Final review — round 2 verified

An independent pass re-verified every fix from the last round landed correctly and re-read the full diff fresh:

1. `test_connection()`/`refresh_manifest()` now go through `HufNativeAdapter._request()` — confirmed no raw `requests.get()` remains anywhere in the diff.
2. `"streaming": false` set, cited against RFC §4.2. `list_agents`/`get_agent_manifest` now call `frappe.has_permission("Agent", "read", ...)`.
3. `test_connection_cmd`/`refresh_manifest_cmd` now check `"write"` permission, with a comment explaining why.
4. `tool_handler.py` checks `frappe.db.exists("DocType", "Delegated Agent Run")` first and raises the existing `RemoteAgentNotImplementedError` — correctly reuses the exception hierarchy rather than inventing a new one.
5. `ws://`/`wss://` rejected explicitly at save time.
6. Both stub ACP-standard adapter classes now carry clear "not yet implemented" docstrings.
7. No regressions — all touched files are net-new additions, no dangling partial edits, no syntax issues.
8. PR description matches the diff exactly, including the honest "still missing" list (the 3 doctypes, stub run endpoints, no frontend).

**Verdict: ready to merge as what it honestly is — safe, tested scaffolding, not a complete feature.** Follow-up work (3 missing DocTypes, real run-lifecycle endpoints, frontend config page) is real net-new scope for a separate PR, not a blocker on this one.

PR #5, opened 2025-11-03
feat:Add perplexity search tool

esafwan:
Important: Merge after merging liteLLM PR.

## Add Perplexity Search Tool

Adds Perplexity Search as a tool for agents via LiteLLM.

### What it does

- Adds "Perplexity Search" to Agent Tool Function types
- Allows agents to search using Perplexity AI
- Uses LiteLLM's search provider for Perplexity integration

### Configuration

`site_config.json`:

```json
{
  "perplexity_api_key": "pplx-..."
}
```

### Get API Key

Get your API key from: https://www.perplexity.ai/settings/api

### Usage

1. Create an `Agent Tool Function` with type "Perplexity Search"
2. Add it to your Agent's tool list
3. The agent can use it to perform searches via Perplexity AI

This follows the same pattern as the Google Search tool implementation.

PR #662, opened 2026-08-24
fix: live-bench QA  — agent config, knowledge embeddings, code execution

esafwan:
## What this is, plainly

Seven small, independent bugfixes found during a live-bench QA sweep of agent config, knowledge/embeddings, execution analytics, and the code-execution tool. Now with test coverage for the three that could be tested without a live Frappe site, and one additional real bug found and fixed while writing those tests.

## The seven fixes

1. **Retry once on MariaDB deadlock** when persisting a background-generated conversation title, instead of dropping the title silently.
2. **Reject invalid JSON** in `Agent.voice_config` at save time instead of persisting it and failing later at voice-session start.
3. **Enforce non-negative `Agent.max_upload_size_mb`** — a negative value was silently breaking every chat upload.
4. **Fix embedding config resolution** for knowledge sources: reads the AI Provider's `api_base_url` (the field that actually exists) instead of a nonexistent `api_base`, and provider-prefixes the model for LiteLLM so local/Ollama-backed embeddings actually reach the provider.
5. **Make the `run_python` code-execution tool reachable end-to-end**: missing import, tool-sync registration, capability-gate widening, and the "Code Execution" type mapping in two consumer modules that lacked one.
6. **Fix a naive/aware datetime comparison crash** in execution analytics when the frontend sends ISO date strings with a timezone suffix.
7. **Add a `before_tests` hook** that clears the site's stale test-record cache, fixing intermittent `ImplicitCommitError` on repeated `bench run-tests` runs.

## What changed since the last review round

The original review found all 7 fixes correct but with zero test coverage. This round closed most of that gap:

- **Added 3 real unit test files**, following this codebase's existing test conventions:
  - Deadlock retry (#1): 4 tests, syntax-checked (couldn't execute — `agent_integration.py` transitively pulls in `litellm`/the OpenAI Agents SDK, too heavy to stub quickly in an isolated clone; should run clean in CI).
  - Embedding config resolution (#4): 7 tests, **executed and passing** against a stub `frappe` module.
  - Analytics datetime fix (#6): 3 tests, **executed and passing** — and verified meaningful by reverting the fix locally, confirming the test fails with the exact `TypeError` the fix addresses, then restoring the fix.
- **Fixed the import-order nit** flagged in review (`import frappe` in `code_execution.py` was in the wrong import group — this also happened to paper over a latent pre-existing bug where the function used bare `frappe.*` calls with no top-level import at all before this PR added one, just misplaced).
- **Found and fixed a real bug while investigating fix #3**: the DocType's `"non_negative": 1` JSON flag alone does *not* enforce server-side — confirmed against this repo's own convention, where `KnowledgeSource` and `ExecutionProfile` both pair `non_negative: 1` with an explicit Python check. Traced the actual impact: without one here, a negative `max_upload_size_mb` reaches the upload-size calculation in `agent_chat.py` and produces a negative byte ceiling, silently breaking every file upload rather than just failing validation. Added the missing `frappe.throw`.

## What's still open

- **Fix #7 (`before_tests` hook) needs a live bench run** to verify — `bench run-tests` behavior isn't something a bare git clone can exercise. Flagging for whoever merges this to do one rerun before/after.
- The deadlock-retry test (#1) should be run in CI rather than trusted on syntax-check alone, given the heavy import chain prevented local execution here.

## Recommendation

Low risk, high confidence. Safe to land once CI confirms the new tests pass and someone does the one manual `bench run-tests` check for fix #7.

*Second-round fixes applied by an automated audit/fix pass over 2026-08-25's revived-branch inventory.*

### DIFF
--- a/huf/ai/agent_integration.py
+++ b/huf/ai/agent_integration.py
@@ -891,6 +891,29 @@ def run_background_summarization(conversation_name, agent_name):
             "Agent Background Summarization Error"
         )
 
+
+
+def _set_conversation_title_with_retry(conversation_name, title, max_retries=2):
+    """Set conversation title, retrying on transient MariaDB deadlocks.
+
+    The title generation runs in a background job that may race with the
+    main request thread updating the same Agent Conversation row. A single
+    short retry is usually enough to land the title without surfacing a
+    non-critical error to the user.
+    """
+    import time
+
+    for attempt in range(max_retries):
+        try:
+            frappe.db.set_value("Agent Conversation", conversation_name, "title", title)
+            return
+        except frappe.QueryDeadlockError:
+            if attempt < max_retries - 1:
+                time.sleep(0.2 * (attempt + 1))
+                continue
+            raise
+
+
 @frappe.whitelist()
 def generate_conversation_title(conversation_name, agent_name):
     """
@@ -929,7 +952,7 @@ def generate_conversation_title(conversation_name, agent_name):
         )
         if title:
             title = title.strip().strip('"').strip("'")
-            frappe.db.set_value("Agent Conversation", conversation_name, "title", title)
+            _set_conversation_title_with_retry(conversation_name, title)
             frappe.db.commit()  # nosemgrep: justified background-job commit
             _emit_conversation_title_updated(conversation_name, title)
     except Exception as e:
--- a/huf/ai/agent_run_analytics_api.py
+++ b/huf/ai/agent_run_analytics_api.py
@@ -28,6 +28,14 @@ def get_execution_analytics(from_date: str | None = None, to_date: str | None = 
         frappe.throw("granularity must be 'hour' or 'day'")
     end = get_datetime(to_date) if to_date else now_datetime()
     start = get_datetime(from_date) if from_date else add_to_date(end, days=-7)
+    # from_date/to_date arrive as ISO strings (often UTC with a Z/offset suffix) from the
+    # frontend, which get_datetime parses as timezone-aware, while now_datetime()/add_to_date()
+    # return naive local datetimes. Mixing the two raises
+    # "can't compare offset-naive and offset-aware datetimes" on the next line.
+    if start.tzinfo is not None:
+        start = start.replace(tzinfo=None)
+    if end.tzinfo is not None:
+        end = end.replace(tzinfo=None)
     if start > end or (end - start).days > MAX_WINDOW_DAYS:
         frappe.throw(f"Date range must be between zero and {MAX_WINDOW_DAYS} days")
     if not frappe.db.exists("DocType", ROLLUP_DOCTYPE):
--- a/huf/ai/flow_tool_executor.py
+++ b/huf/ai/flow_tool_executor.py
@@ -183,6 +183,7 @@ def _resolve_function_path(tool_doc: dict) -> str | None:
 		"Set Conversation Data": "huf.ai.sdk_tools.handle_set_conversation_data",
 		"Load Conversation Data": "huf.ai.sdk_tools.handle_load_conversation_data",
 		"Perplexity Search": "huf.ai.tools.perplexity.handle_perplexity_search",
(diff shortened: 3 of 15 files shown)


esafwan:
## Automated review — solid fixes, needs test coverage before merge

All 7 fixes were verified line-by-line against current develop (not just the branch's original diff) to rule out stale reversions. Summary:

| # | Fix | Verdict |
|---|-----|---------|
| 1 | MariaDB deadlock retry on title persistence | Correct — exactly one retry, no infinite loop, exhausted-retry path matches prior behavior |
| 2 | Reject invalid `voice_config` JSON | Correct — wired into `validate()`, catches `TypeError`/`ValueError`, `json` already imported |
| 3 | Non-negative `max_upload_size_mb` | Correct approach (DocType-level constraint) but **unverified** — confirm Frappe's Int-field validation actually fires on save with a negative value |
| 4 | Embedding config `api_base_url` fix | Correct, and confirmed **still needed** — the old `api_base` attribute genuinely doesn't exist on `AI Provider`, so this was silently `None` before. Also fixes a real url+port double-append bug |
| 5 | `run_python` reachable end-to-end | Correct, and the capability-gate change **narrows by identity, doesn't widen scope** — verified it mirrors the existing SSH/Docker gate pattern exactly. Confirmed `code_execution.py` genuinely used bare `frappe.*` without importing `frappe` — real `NameError` bug fixed. Nit: `import frappe` is out of alphabetical order, will likely trip isort/ruff |
| 6 | Naive/aware datetime crash in analytics | Correct — confirmed `get_datetime()` vs `now_datetime()` tz mismatch is real. Stripping tzinfo rather than converting is the right tradeoff here (low-impact edge case at day boundaries, only affects a window-length guard) |
| 7 | `before_tests` cache-clear hook | Completes a stub that was never wired up (not re-enabling something disabled for a reason — confirmed no commit evidence of deliberate removal). Guards look defensive and correct |

**Gap: zero test coverage.** No test files are touched anywhere in this diff — all 12 files are production code. At minimum #1, #4, and #6 are cheaply unit-testable and should get coverage before this lands. #3 and #7 both deserve a quick manual verification pass (negative-value save; a repeated `bench run-tests` run against a shared site) rather than just code review.

**Verdict: needs changes before merge** — not because anything is wrong, but for test coverage on the easy ones and manual verification on #3/#7. The batch is fine to land together as one PR, no need to split it.

esafwan:
## Final review — round 2 verified

An independent pass re-verified the round-2 additions and re-read the full 12-file diff fresh:

1. All 3 new test files are real, meaningful tests with exact assertions (call counts, resolved config values, retry/exhaustion behavior) — not smoke tests. The analytics-tz test (fix #6) is the thinnest of the three (mostly "does not raise"), but its third case does directly target the mixed aware/naive regression by construction. Acceptable.
2. `import frappe` ordering fixed correctly. (Noted, out of scope: a pre-existing stray import placed after a function definition in the same file — predates this PR, untouched by it.)
3. The new `frappe.throw` for negative `max_upload_size_mb` is correct — right field, right comparison, reachable on every save. Verified against the real consumer in `agent_chat.py`: a negative value is truthy in Python, so the `or 25` fallback doesn't rescue it, confirming both the bug and the fix are accurately described.
4. Full diff read end to end — internally consistent, no leftover debug code or unresolved TODOs.
5. PR description matches the diff precisely, including the nuanced detail about registry-synced tools landing as `"App Provided"` rather than `"Code Execution"` (exactly why fix #5's gate needed widening).
6. Fresh look at fix #5's capability-gate change: **it's a tightening, not a loosening.** Registry-synced tools are stamped `types = "App Provided"`, so the *original* narrow gate would have let a registry-synced `run_python` bypass capability checks entirely — the widened gate closes that gap and mirrors the existing SSH/Docker gate pattern already in the same file.

**Verdict: ready to merge — needs only the one still-flagged manual `bench run-tests` check** for fix #7's `before_tests` hook, which the PR description already notes can't be verified from a bare clone.

PR #663, opened 2026-08-24
feat: community OpenAI/Kimi subscription adapters (revived, conflict-resolved)

esafwan:
## What this revives

Orphan branch `feat/community-subscription-adapters` (never had a PR, last commit 2026-08-04) forward-ported onto current `develop`. **Not a clean cherry-pick** — 3 weeks of independent divergence on both sides required manual, additive conflict resolution (see below). Flagging that explicitly since this one needs closer review than a clean rebase would.

## What it adds

Lets a user attach their **personal OpenAI or Kimi subscription** to Huf instead of pasting a raw API key — an OAuth-style "community subscription" connection flow.

- `huf/ai/providers/adapters/openai.py`, `openai_community.py` — new provider adapters
- `huf/huf/doctype/ai_provider_connection/` — new DocType for the subscription-linking connection
- `huf/www/sub_oauth.py`, `sub_oauth.html` — the OAuth landing pages
- A "Connections" UI entry point in `AiProvidersHeaderActions.tsx`
- Confirmed via grep against `develop`: none of `openai_community`, `sub_oauth`, or `ai_provider_connection` exist anywhere in current develop — genuinely new capability.

## Conflict resolution — read this before merging

Two files needed manual merging because develop grew independently in the same spots since Aug 4:

1. **`huf/ai/agent_integration.py`** — develop had separately added a `project: str = None` param to `run_agent_stream` in the same location this branch adds `subscription_connection_name: str = None`. Resolved by keeping both params (additive, no collision in behavior).
2. **`huf/huf/doctype/ai_provider/ai_provider.json`** — both sides independently extended the `provider_brand` Select options list (develop added `moonshot, ai21, baseten, cloudflare-workers-ai, clarifai, nomic, replicate, sagemaker, stability-ai, vllm, watsonx`; this branch added `openai_community, kimi_community`). Resolved by unioning both lists — no options were dropped from either side.
3. **`frontend/src/components/AiProvidersHeaderActions.tsx`** — develop had a plain "Add provider" button; this branch wraps it with a new "Connections" nav button. Kept the incoming version (adds Connections, preserves Add Provider).

All resolutions were verified additive: nothing currently in develop was deleted or regressed. Python files pass `py_compile`; both touched doctype JSONs parse as valid JSON.

## Before merging

- [ ] **This one needs a real review of the conflict resolutions above**, not just the original diff — an automated agent resolved them additively, but a human should confirm the `ai_provider.json` option-list merge and the `run_agent_stream` param addition don't have subtler interaction effects
- [ ] OAuth landing page (`sub_oauth.py`/`.html`) handles user-facing subscription credentials — worth a security pass on how tokens/secrets are stored and whether they follow Huf's existing `Integration Credential` conventions
- [ ] This is a **draft** — original author (esafwan) should confirm this is still the intended UX before it goes further, since 3 weeks have passed since the branch was written

*Opened as part of an automated audit/revival pass over 2026-08-25's orphan-branch inventory.*

### DIFF
--- a/HOW_TO_TEST.md
+++ b/HOW_TO_TEST.md
@@ -1,0 +1,280 @@
+# How to Test Community Subscription Adapters
+
+> Branch: `feat/community-subscription-adapters`  
+> Container: `fdocker_devcontainer-frappe-1`  
+> Bench path (inside container): `/workspace/development/16`  
+> Site: `huf.localhost`  
+> Host URL: `http://huf.localhost:8100`
+
+## What this branch adds
+
+- A new **AI Provider Connection** DocType that stores per-user OAuth/device-code tokens.
+- A subscription-adapter registry under `huf/ai/providers/adapters/`.
+- Community adapters that reuse public client credentials from popular AI CLI plugins:
+  - **OpenAI Community** (`openai_community_subscription`) — OAuth PKCE, manual paste flow.
+  - **Kimi For Coding** (`kimi_community_subscription`) — OAuth device flow.
+  - **Mock** (`mock_subscription`) — for CI / offline regression tests.
+- An OAuth callback page at `/huf/sub_oauth`.
+- Routing changes in `huf/ai/run.py` and `huf/ai/agent_integration.py` so agents can run through a subscription connection even when the shared AI Provider has no API key.
+
+> ⚠️ These adapters use **community-reversed OAuth client IDs**. They are not official integrations. Enable them only for experimentation; using them with real accounts may violate the provider’s Terms of Service.
+
+---
+
+## 1. Environment setup
+
+Run all bench commands through the Frappe container:
+
+```bash
+docker exec fdocker_devcontainer-frappe-1 bash -c "cd /workspace/development/16 && bench --site huf.localhost <COMMAND>"
+```
+
+### 1.1 Migrate the site
+
+```bash
+docker exec fdocker_devcontainer-frappe-1 bash -c "cd /workspace/development/16 && bench --site huf.localhost migrate"
+```
+
+### 1.2 Enable the adapters
+
+```bash
+docker exec fdocker_devcontainer-frappe-1 bash -c "cd /workspace/development/16 && \
+  bench --site huf.localhost set-config enable_openai_community_subscription_adapter 1 && \
+  bench --site huf.localhost set-config enable_kimi_community_subscription_adapter 1 && \
+  bench --site huf.localhost set-config enable_mock_subscription_adapter 1"
+```
+
+### 1.3 Verify the callback page
+
+From your Mac:
+
+```bash
+curl -s -o /dev/null -w '%{http_code}' http://huf.localhost:8100/huf/sub_oauth
+# Expected: 200
+```
+
+---
+
+## 2. Run the test suite
+
+```bash
+docker exec fdocker_devcontainer-frappe-1 bash -c "cd /workspace/development/16 && \
+  bench --site huf.localhost run-tests --module huf.ai.tests.test_subscription_adapter_mock && \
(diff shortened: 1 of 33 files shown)


esafwan:
## Automated review — real multi-tenant data-exposure gaps, needs a security pass before merge

### Conflict resolutions (the parts done by the revival agent)
- `ai_provider.json` `provider_brand` union: verified programmatically — 38 options, all unique, valid JSON. **Clean.**
- `run_agent_stream`'s `subscription_connection_name` + independently-added `project` param: no collision, additive. **Minor gap**: `run_agent_sync`'s two context-dict builds add `subscription_connection_name` but not `"user": frappe.session.user` the way `_execute_agent_run`/`run_agent_stream` do — inconsistent, will bite once `run_agent_sync` runs from a background-worker context where ambient session isn't reliable. Add `"user": frappe.session.user` there for parity.

### Major — OAuth callback has no ownership/session check
`huf/www/sub_oauth.py`'s `get_context` resolves the pending connection purely by matching the state token against **all** `AI Provider Connection` rows system-wide, then saves with `ignore_permissions=True` — no check that `frappe.session.user == connection_doc.user`, and no login requirement on the page at all. The 128-bit state token provides some protection, but there's no defense-in-depth: an unauthenticated visitor with a leaked `code`/`state` pair could complete another user's OAuth authorization. **Fix: verify session user matches connection owner, and require login on this page.**

### Major/borderline blocker — AI Provider Connection permissions leak metadata across users
The permission block was copy-pasted from `ai_provider.json` (an org-wide config doctype), which is wrong here — this doctype is per-user (has a `user` Link field), but the "Huf User" role gets `read: 1` with **no `if_owner` restriction**. Any user with that role can list every other user's connection rows (`connection_name`, `user`, `provider`, `account_email`, `account_id`, `auth_status`) via direct doctype access (`/api/resource/AI Provider Connection`) — bypassing the `_can_access_connection()` gate that the whitelisted RPC methods correctly use. Password-typed fields (`access_token`/`refresh_token`) are still masked by Frappe's own field-level protection, so this is metadata/PII exposure (which OpenAI account is linked to which Huf user), not a raw token leak — but it's still a real tenancy bug. **Fix: add `if_owner: 1` to the Huf User permission row.**

### Minor / worth a look
- `openai_community.py` hardcodes a reverse-engineered OAuth client ID from a community project that impersonates the Codex CLI — gated behind an opt-in site config flag (good), but worth flagging given the likely ToS implications.
- Both adapters report `cost: 0.0` unconditionally — correct in that subscriptions aren't billed to the org, but if Huf's usage-metering pipeline relies on the `cost` field for per-user rate/usage caps, subscription-routed runs are invisible to it. Worth confirming with whoever owns metering.
- No rate limiting in either adapter — a Huf agent loop could hammer a user's personal ChatGPT session limits hard enough to get their account flagged. The user's own risk, but worth a TODO.
- `openai.py`/`openai_community.py` duplicate ~150 lines of PKCE/JWT/SSE parsing — maintainability nit, could share a mixin.
- No test exercises `sub_oauth.py`, the doctype permission surface, or the `_get_subscription_connection` auto-discovery precedence — the two real findings above would not be caught by current tests.

### Good
- Credential storage (`access_token`/`refresh_token`/`auth_payload` as Frappe `Password` fields) matches the house pattern used by `integration_credential`/`ai_provider` — not rolled-own.
- `matches_model()`/`eligible_models` allowlist is a real (if opt-in) scoping mechanism, respected by `RunProvider`'s auto-discovery — not cosmetic.
- Adapter test coverage for the auth/refresh/token-storage contract itself is solid (mock, kimi_community, openai_community all covered).

**Verdict: needs a deeper security review before merge** — specifically the OAuth ownership check and the doctype permission scoping. The conflict-resolved sections themselves check out clean.

PR #694, opened 2026-09-04
fix(perms): add per-run owner checks to runtime doctypes via has_permission hooks

esafwan:
## Summary
Implements WP-01 of the AgentSecurityRemediation track: record-level authorization for runtime doctypes.

**Findings closed:** F-01 (fully), F-06 (fully), F-31 (fully), F-32 (fully), F-14 (partially — ST-01.8: `frappe.set_user` for API-key requests; credential-scope enforcement stays out of scope)

## What this does
- Adds `has_permission` hooks in `huf/ai/hooks.py` for Agent Run, Agent Message, Agent Conversation, Agent Tool Call, and Agent Context Artifact, restricting single-document reads to the owner, System Manager, or a capability holder (`agent.view_all` / `chat.view_all`).
- `huf/ai/record_access.py` centralizes the owner-check logic reused by both the `has_permission` hooks and (in WP-02) list scoping.
- `huf/api/v1/router.py`: `ApiV1Router.render` now calls `frappe.set_user(context.user)` for `AuthMode.API_KEY` requests so records created via the API-key auth path are correctly attributed (closing the F-14 user-switch gap).
- Adds `get_agent_run_status`/`get_run_context_metrics` record-level checks in `huf/ai/agent_integration.py`.

- A real bug: streaming API-key requests lost the correct `frappe.session.user` before the lazy stream generator executed — fixed.
- `test_router_set_user.py` originally tested a re-implementation of the router logic rather than the router itself — rewrote to exercise the real code path.
- Converted several new test files from plain pytest-style functions to `IntegrationTestCase` classes — `bench run-tests` only discovers unittest-based tests, so the originals silently ran 0 tests.
- `frappe.Document.insert()` always stamps `owner` to the current session user, overriding a test's explicit `doc.owner = "alice@example.com"` assignment before insert — fixed via `doc.db_set("owner", ..., update_modified=False)` after insert across the new record-auth tests.
- Added missing required fixtures (`Test Agent`, `session_id` on Agent Conversation) so the new tests pass on a fresh bench.
- `has_permission` hooks can only narrow an already-granted role permission, never grant a new one on their own (Frappe's `has_controller_permissions` contract) — the two tests going through the full `Document.check_permission()` path needed a real `Huf User`-roled test user, not just a made-up email.

## Test plan
- [x] `bench migrate` clean on a fresh bench restored from a blank reference site.
- [x] All 8 test modules for this WP pass under `bench --site <site> run-tests --app huf --module <module>`: `test_agent_run_record_auth`, `test_agent_conversation_record_auth`, `test_agent_context_artifact_record_auth`, `test_agent_tool_call_record_auth`, `test_agent_message_record_auth`, `test_get_agent_run_status_perms`, `test_get_run_context_metrics_perms`, `test_hooks_registration`, `test_router_set_user`.

### DIFF
--- a/huf/ai/agent_integration.py
+++ b/huf/ai/agent_integration.py
@@ -1429,7 +1429,7 @@ def get_agent_run_status(agent_run_id: str):
     run = frappe.db.get_value(
         "Agent Run",
         agent_run_id,
-        ["name", "agent", "status", "response", "error_message", "conversation"],
+        ["name", "agent", "status", "response", "error_message", "conversation", "owner"],
         as_dict=True,
     )
     if not run:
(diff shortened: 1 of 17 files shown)


esafwan:

PR #695, opened 2026-09-04
fix: convert secret fields from Data to Password with re-encrypting patch

esafwan:
## Summary
Implements WP-03 of the AgentSecurityRemediation track: secrets out of Data fields.

**Findings closed:** F-03 (fully), F-05 (partially — storage half; the `initiating_user` half is WP-06)

## What this does
- Converts MCP header values, tool HTTP header values, and both Automation Trigger webhook-key fields from `Data` to `Password`, with a re-encrypting migration patch (`huf/patches.txt`).
- Every call site that reads these secrets is updated to use `get_password()` instead of direct field access.
- Drops `report`/`export` permissions from Huf User and Huf Viewer rows on the affected doctypes.

- `frappe.utils.get_random_string` doesn't exist — the real API is `frappe.utils.random_string(length)`.
- The new test file's fixture helpers were missing several mandatory fields (`Automation.agent`/`instruction`, `MCP Server.server_url`/real `transport_type`, `Agent Tool Type`'s actual autoname field `name1`, `Agent Tool Function.description`).
- The masking test used `set_encrypted_password()` directly and expected the main-table column to already hold the `"*"*len` dummy mask — but that masking only happens inside `Document._save_passwords()` on a real `doc.save()`, not from calling `set_encrypted_password()` on its own. Rewrote those tests to assign the plaintext value and save the document, matching how the real code path works.
- Removed a `with patch("frappe.session")` wrapper that fully replaced the session mock and broke fixture creation without actually testing anything relevant.

## Test plan
- [x] `bench migrate` clean (patch applies without error).
- [x] `test_secrets_migration.py` — all 6 tests pass.

### DIFF
--- a/huf/ai/automation_webhook.py
+++ b/huf/ai/automation_webhook.py
@@ -84,7 +84,7 @@ def handle_automation_webhook():
 	trigger = frappe.get_doc("Automation Trigger", trigger_name)
 
 	supplied_key = frappe.get_request_header("X-Webhook-Key") or ""
-	expected_key = trigger.webhook_key or ""
+	expected_key = trigger.get_password("webhook_key") or ""
 	if not expected_key or not hmac.compare_digest(supplied_key, expected_key):
 		frappe.local.response["http_status_code"] = 401
 		return dict(_INVALID_KEY)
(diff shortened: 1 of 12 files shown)


esafwan:

PR #696, opened 2026-09-04
fix: fail-closed gateway signature verification

esafwan:
## Summary
Implements WP-04 of the AgentSecurityRemediation track: fail-closed gateway verification.

**Findings closed:** F-04 (fully), F-18 (fully), F-35 (partially — rate limiting)

## What this does
- Makes signature verification mandatory and fail-closed in every gateway adapter (WhatsApp, Telegram, Messenger, Instagram, Slack, Email, Google Chat, Microsoft Teams), modelled on the existing Discord adapter.
- A migration patch disables gateways with unset secrets instead of silently accepting unverified traffic.
- Constrains `Gateway.execution_user` to an allow-listed role.

## Deviations / fixes found and fixed during this review pass
- **Real security bug found and fixed**: the Microsoft Teams adapter was the one adapter that still failed *open* — `verify_inbound()` returned `True` unconditionally whenever `app_id` wasn't configured ("backwards compatibility"), which is exactly the behavior this WP exists to eliminate. Every other adapter validates required credentials in `__init__` and refuses to construct at all when they're missing; Teams now does the same (`credential_schema.missing_required()` in `__init__`, raising `ValueError`), and `verify_inbound()` no longer has an "unconfigured, skip the check" branch.
- The conformance test suite (`test_gateway_conformance.py`) assumed adapter construction always succeeds and only `verify_inbound()` can fail closed — but several adapters (correctly) fail closed by raising `ValueError` at construction time instead. Updated the test to accept either form.
- The test suite referenced a `"sms"` gateway provider that isn't actually registered (`KeyError` from `_ADAPTER_LOCATIONS`) — removed those references.
- A tuple-unpack bug in the SMS gateway credential check migration was fixed during the earlier automated review pass (see commit history).

## Test plan
- [x] `bench migrate` clean.
- [x] `test_gateway_conformance.py` — all 9 tests pass (covering whatsapp, telegram, messenger, instagram, email, google_chat, microsoft_teams, slack, plus the cross-provider "never raises from verify_inbound" conformance check).

### DIFF
--- a/huf/ai/gateway_adapters/email.py
+++ b/huf/ai/gateway_adapters/email.py
@@ -25,7 +25,7 @@ class EmailGatewayAdapter(GatewayAdapter):
 	provider_id = "email"
 	credential_schema = GatewayCredentialSchema(
 		(
-			GatewayCredentialField("webhook_secret", "Webhook Verification Secret (Optional)", required=False),
+			GatewayCredentialField("webhook_secret", "Webhook Verification Secret", required=True),
 			GatewayCredentialField("sender_email", "Default Outbound Sender Email (Optional)", required=False),
 		)
 	)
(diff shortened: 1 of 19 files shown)


esafwan:

PR #697, opened 2026-09-04
sec: validate App Provided tool paths against hook registry, gate fetch_tool_parameters_from_code

esafwan:
## Summary
Implements WP-05 of the AgentSecurityRemediation track: allow-set guard for dynamic tool function paths.

**Findings closed:** F-07 (fully), F-19 (partially — only the `fetch_tool_parameters_from_code` slice; the other 8 endpoints in F-19 are WP-06)

## What this does
- `App Provided` tool function paths are now validated against the hook-declared tool registry (`get_hook_declared_function_paths()`), closing the F-07 RCE vector.
- `is_whitelisted()` remains the gate for `Custom Function` only.
- `fetch_tool_parameters_from_code` now requires the `Huf Manager` role (`frappe.only_for`).

- A test used `"subprocess.getoutput"` as a non-whitelisted `Custom Function` path expecting `frappe.PermissionError`, but `frappe.get_attr()` raises `AppNotInstalledError` (a `ValidationError` subclass) before `is_whitelisted()` is ever reached, since `subprocess` isn't an installed Frappe app prefix. Switched to a real, resolvable-but-never-whitelisted frappe function so the test actually exercises `is_whitelisted()`.
- `frappe.only_for()` unconditionally no-ops when `frappe.flags.in_test` is set — which `bench run-tests` sets for the entire run — so the `fetch_tool_parameters_from_code` authorization test could never actually exercise the real gate. Fixed by temporarily clearing the flag around the calls under test and creating real test users with the roles being asserted.
- A test mocked `huf.ai.sdk_tools.get_hook_declared_function_paths`, but that function is imported locally inside `sdk_tools.get_function_from_name` from `huf.ai.tool_registry` — not a module-level attribute of `sdk_tools`. Patched the real source module instead.

## Test plan
- [x] `bench migrate` clean.
- [x] All 3 test modules pass: `test_agent_tool_function_validation` (3 tests), `test_fetch_tool_parameters_authorization` (1 test), `test_tool_invocation_runtime` (3 tests).

### DIFF
--- a/huf/ai/sdk_tools.py
+++ b/huf/ai/sdk_tools.py
@@ -482,7 +482,7 @@ def create_function_tool(
         FunctionTool: Function tool
     """
 
-    function = get_function_from_name(tool_name)
+    function = get_function_from_name(tool_name, tool_type=tool_type)
 
     if not function:
         return None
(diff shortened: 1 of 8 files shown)


esafwan:

PR #698, opened 2026-09-04
huf/ai: require agent.use capability and harden HTTP tool handler SSRF/credential controls

esafwan:
## Summary
Implements WP-07 of the AgentSecurityRemediation track: HTTP tool handler hardening.

**Findings closed:** F-08 (fully)

## What this does
- Requires the `agent.use` capability on `handle_http_request`/`handle_get_request`/`handle_post_request` (guest sessions remain exempt where explicitly allowed per-tool).
- Binds tool-defined headers (including credentials) to the tool's own configured origin; headers are stripped on cross-origin requests and across redirects.
- Streams the response body with a hard size cap (10MB), including a Content-Length fast-path rejection.
- Adds per-user rate limiting to the HTTP handlers.

Note for downstream consumers: WP-06 reuses `validate_url` (in this file) for its own SSRF checks (`resolve_mcp_connection`), so this branch should land/rebase-order before WP-06 — already reflected in this track's branch stack (`fix/maintenance-endpoint-auth` is built on top of this branch).

- **Real bug found and fixed**: Authorization-class header stripping on cross-origin requests was case-sensitive, so a header literally named e.g. `authorization` (lowercase) would survive a cross-origin redirect intact — a credential-leak gap in exactly the mechanism this WP exists to harden. Fixed to strip case-insensitively.
- Several tests in `test_http_handler_access.py` relied on real DNS resolution of `api.example.com` via `validate_url`'s SSRF check — not guaranteed even with real network access (subdomains of `example.com` aren't reserved to resolve), and this bench's container has no outbound DNS at all. Mocked `validate_url` consistently with the tests that already got this right.
- Two tests left `frappe.get_doc` unmocked or `http_headers` as an unconfigured `Mock` (which is truthy and non-iterable), causing unrelated crashes before the behavior under test was ever reached. Fixed to match the established mocking pattern used elsewhere in the same file.

## Test plan
- [x] `bench migrate` clean.
- [x] `test_http_handler_access.py` — all 21 tests pass (authorization, guest gating, header origin binding + redirect stripping, response size capping, rate-limit decorator presence).

### DIFF
--- a/huf/ai/http_handler.py
+++ b/huf/ai/http_handler.py
@@ -64,6 +64,18 @@ def _is_public_ip(ip_str: str) -> bool:
 	return True
 
 
+def extract_origin(url_str):
+	"""
+	Extract (scheme, host, port) from a URL for origin matching.
+
+	Returns a tuple (scheme, hostname, port) where port is derived from
+	the URL's port attribute or defaults to 443 for https, 80 for http.
+	"""
+	parsed = urlparse(url_str)
+	port = parsed.port or (443 if parsed.scheme == "https" else 80)
+	return (parsed.scheme, parsed.hostname, port)
+
+
 def validate_url(url):
 	"""
 	Validate URL to prevent SSRF and other attacks.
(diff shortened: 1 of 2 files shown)


esafwan:

PR #700, opened 2026-09-04
fix(perms): add list-scope PQC hooks for Agent Tool Call, API Key, snapshots and procedures

esafwan:
## Summary
Implements WP-02 of the AgentSecurityRemediation track: list scoping for unscoped doctypes.

**Stacking note:** this branch is built on top of `fix/record-authz` (#694, WP-01), reusing its `has_permission`/`record_access` design for consistent list vs. single-doc scoping. This PR's diff includes WP-01's commits until #694 merges — that's expected; only the last handful of commits (the `permission_query_conditions` hooks and their tests) are WP-02's own.

**Findings closed:** F-02 (fully), F-29 (fully), plus Agent Procedure Run scope verification

## What this does
- Adds `permission_query_conditions` hooks for Agent Tool Call, Agent Run Prompt Snapshot, Huf API Key, and Agent Procedure Run, removing the site-wide enumeration primitive (F-02's core issue: any authenticated user could list every run/tool-call/API-key on the site).
- List scoping reuses the owner-check logic from WP-01's `huf/ai/record_access.py` for consistency between list and single-doc paths.

- Same pytest-style-to-unittest conversion issue as WP-01, for this WP's own 4 new test files.
- `Huf API Key`'s `before_insert()` requires a transient `_raw_secret` attribute (raises otherwise) and `insert()` resets `owner` to the session user regardless of what the test set beforehand — fixed the test fixtures.
- `Agent Procedure Run`'s `procedure` field links to the `Agent Procedure` doctype (not a nonexistent `"Procedure"` doctype), and its `status` field uses Title Case values (`"Completed"`, not `"success"`).
- The `Agent Run Prompt Snapshot` list-scope test initially used the `Huf Manager` role for the "should be scoped" test cases — but `Huf Manager` legitimately carries the `agent.view_all` capability by design, so it always bypasses scoping. Switched to `Huf User`, and since that role has no DocPerm read access at all on this doctype by default, granted it via a `Custom DocPerm` (without `agent.view_all`) so the scoping condition is actually exercised.

## Test plan
- [x] `bench migrate` clean.
- [x] All 4 of this WP's own test files pass (19 tests): `test_api_key_list_scope`, `test_procedure_run_list_scope`, `test_agent_run_prompt_snapshot_list_scope`, `test_agent_tool_call_list_scope`.
- [x] `test_hooks_registration.py` (shared with WP-01, extended here) — both tests pass.


esafwan:

PR #701, opened 2026-09-04
Stability: defer Agent planning and scheduled runs from request to queue

esafwan:
## Summary
Implements WP-08 of the AgentSecurityRemediation track: takes the LLM loop out of the web request.

**Stacking note:** this branch is built on top of `fix/record-authz` (#694, WP-01) — it depends on ST-01.3's guest-polling narrowing of `get_agent_run_status`. This PR's diff includes WP-01's commits until #694 merges; only the last several commits (planning/scheduling deferral) are WP-08's own.

**Findings closed:** F-10 (fully), F-11 (fully), F-25 (partially)

## What this does
- Flips `Agent.run_immediately` to default `0` (queue by default; existing agents keep their stored value) and enqueues default-plan generation from Agent save instead of running it inline in the request.
- The scheduler now pre-claims `next_execution` via a conditional `UPDATE ... WHERE` (checking `ROW_COUNT()`) before enqueuing, so two overlapping scheduler ticks can't both dispatch the same trigger.

- **Real bug found and fixed**: the scheduler's dedup pre-claim read `ROW_COUNT()` *after* `COMMIT`, which resets it to 0 on MySQL/MariaDB — so the row-count check always saw 0 and no scheduled agent would ever actually run. Fixed to read it before commit.
- An `after_insert()` path was still blocking on the LLM loop instead of deferring to the queue like `on_update()` does — fixed.
- An orphaned migration patch (`log_run_immediately_agents`) wasn't registered in `huf/patches.txt` — fixed.
- The new `test_agent_planning.py` had several bugs: `frappe.get_doc(dict)` doesn't apply doctype-level field defaults (only `frappe.new_doc()` does) — a test asserting `run_immediately == 0` by default was actually asserting on `None`; `provider`/`model` are real Link-field fixtures (`AI Provider`/`AI Model`), not free-text strings a document controller happens to accept without validation; and `frappe.session.user` can't be `mock.patch()`'d as a dotted string path (`frappe.session` doesn't expose a normal `__dict__` outside a request context) — replaced with direct assignment/restore in `setUp`/`tearDown`.

## Test plan
- [x] `bench migrate` clean.
- [x] `test_agent_planning.py` — all 8 tests pass (run_immediately default, planning enqueue behavior, scheduler pre-claim conditional UPDATE and rowcount handling).

### DIFF
--- a/huf/ai/agent_integration.py
+++ b/huf/ai/agent_integration.py
@@ -1429,7 +1429,7 @@ def get_agent_run_status(agent_run_id: str):
     run = frappe.db.get_value(
         "Agent Run",
         agent_run_id,
-        ["name", "agent", "status", "response", "error_message", "conversation"],
+        ["name", "agent", "status", "response", "error_message", "conversation", "owner"],
         as_dict=True,
     )
     if not run:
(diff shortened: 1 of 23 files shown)


esafwan:

PR #702, opened 2026-09-04
huf/ai: require authorization on maintenance and tool endpoints

esafwan:
## Summary
Implements WP-06 of the AgentSecurityRemediation track: authorization on maintenance and tool endpoints.

**Stacking note:** this branch is built on top of `fix/http-handler-hardening` (#698, WP-07) — it reuses WP-07's hardened `http_handler.validate_url` for `resolve_mcp_connection`'s own SSRF checks, per the plan's explicit reuse requirement. This PR's diff includes WP-07's commits until #698 merges; only the last few commits are WP-06's own.

**Findings closed:** F-19 (fully except `fetch_tool_parameters_from_code`, which is WP-05/#697), F-09 (fully), F-05 (partially: closed on the `trigger_app_event` path by passing `initiating_user`; the webhook path was already closed by the existing owner-bound `_check_run_as_user_permission` guard in `run_automation` — confirmed by test, no code change needed there), F-20 (fully, realtime scoping)

## What this does
- Adds the missing authorization check to each endpoint identified in F-19 (maintenance/tool endpoints that were whitelisted without any permission gate).
- Removes the `@frappe.whitelist()` decorator from `poll_pending_batch_jobs` and the scheduler entrypoints (internal-only, never meant to be externally callable).
- Scopes `client_side_tool.py`'s `publish_realtime` delivery to the conversation owner rather than broadcasting broadly.
- `resolve_mcp_connection` reuses `http_handler.validate_url` instead of duplicating SSRF logic.

- **Real bug found and fixed**: `client_side_function`'s `_, raw_payload = popped` throwaway-variable assignment shadows the module-level `from frappe import _` translation-function import for the *entire function* (Python's scoping rule: any assignment anywhere in a function body makes that name local throughout), causing `UnboundLocalError` on the earlier `frappe.throw(_("Not permitted"), ...)` call — i.e. the permission-denied path itself was broken and would crash instead of cleanly rejecting. Renamed the throwaway variable.
- A missing `frappe._` import in `orchestrator.py` and an SSRF-check duplication (vs. reuse of WP-07's `validate_url`) were fixed during the earlier automated review pass (see commit history).
- Two tests in `test_maintenance_endpoints.py` needed `frappe.get_doc` mocked (pure-mock convention) and `mock_throw.side_effect` set so the mocked `frappe.throw` actually raises and halts execution — without it, tests silently fell through into code paths they weren't meant to reach.

## Test plan
- [x] `bench migrate` clean.
- [x] `test_maintenance_endpoints.py` — all 19 tests pass, covering every ST-06.x sub-task's permission/scoping assertions.

### DIFF
--- a/huf/ai/automation_app_event.py
+++ b/huf/ai/automation_app_event.py
@@ -49,6 +49,7 @@ def trigger_app_event(app_name: str, event_name: str, payload: dict | None = Non
 					"event_name": event_name,
 					"payload": payload or {},
 				},
+				initiating_user=frappe.session.user,
 			)
 			triggered_automations.append(t["automation"])
 		except Exception:
--- a/huf/ai/automation_scheduler.py
+++ b/huf/ai/automation_scheduler.py
@@ -12,7 +12,6 @@ _LOCK_PREFIX = "huf:automation_scheduler:lock:"
 _LOCK_TTL_SEC = 60
 
 
-@frappe.whitelist()
 def run_due_automations():
 	"""Fire every due Schedule-type Automation Trigger.
 
--- a/huf/ai/batch_poll.py
+++ b/huf/ai/batch_poll.py
@@ -332,7 +332,6 @@ def _process_batch_job(job: dict) -> None:
 	frappe.db.commit()  # nosemgrep: justified background-job commit
 
 
-@frappe.whitelist()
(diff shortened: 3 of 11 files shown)


esafwan:

PR #703, opened 2026-09-04
Stability & security: add unified run budget (deadline, max_turns ceiling, depth, spend cap)

esafwan:
## Summary
Implements WP-09 of the AgentSecurityRemediation track: one shared run budget.

**Stacking note:** this branch is built on top of `fix/agent-run-dequeue` (#701, WP-08), which is itself on top of `fix/record-authz` (#694, WP-01). This PR's diff includes both branches' commits until they merge; only the last several commits are WP-09's own.

**Findings closed:** F-12 (fully), F-13 (fully — see note below on the doc-event enforcement vector), F-21 (fully)

## What this does
- Introduces one shared `RunBudget` primitive (`huf/ai/run_budget.py`): wall-clock deadline per round, server-side `max_turns` ceiling, recursion depth counter (fail-closed: `current_depth >= max_depth` raises — reaching the ceiling is blocked, not just exceeding it), and a cumulative per-chain spend cap.
- Persists the budget on `Agent Run` and rebuilds it via `RunBudget.from_run_doc()` at each worker/hook entry point (queue boundary), so it survives across enqueue/dequeue.
- A new `Agent Allowed Model` child doctype backs a model/provider override allowlist (`Agent Settings`), enforced fail-closed unless the caller holds an explicit capability escape hatch.

**On the doc-event depth-check vector** flagged in this track's own review notes (`_FINAL_RECHECK.md` item 11): confirmed during this review pass that it is in fact handled — the doc-event path now has an equivalent mechanism reading the budget from the run document rather than a contextvar, so F-13 is closed on both the enqueue-boundary and doc-event vectors, not just the former as the recheck doc anticipated.

- **Real bug found and fixed**: `Agent Allowed Model`'s doctype JSON existed but was missing its Python controller module and `__init__.py` entirely, which broke `bench migrate` outright with `ModuleNotFoundError`.
- Multiple test bugs across `test_run_budget.py`, `test_run_budget_enforcement.py`, and `test_spend_cap.py`: unittest's `assertRaises(...) as ctx` exposes the exception as `ctx.exception`, not pytest's `ctx.value` (leftover from an earlier pytest-to-unittest conversion pass); a test asserting "depth at ceiling succeeds" contradicted `check_depth`'s actual (and its own docstring's intended) fail-closed `>=` semantics — fixed the test to match; and a test's own arithmetic (`30+40+20+5=95`) never actually exceeded its 100 spend cap, so the expected `ValidationError` never fired — added a genuine over-cap push.
- Converted this WP's own 7 new pytest-style test files to `IntegrationTestCase`/proper unittest classes (same systemic issue as WP-01/WP-02).

## Test plan
- [x] `bench migrate` clean.
- [x] All 7 of this WP's own test files pass: `test_agent_settings` (7), `test_litellm_budget` (4), `test_model_override_allowlist` (10), `test_recursion_depth` (11), `test_run_budget` (10), `test_spend_cap` (12), `test_run_budget_enforcement` (14).

### DIFF
--- a/huf/ai/agent_integration.py
+++ b/huf/ai/agent_integration.py
@@ -1429,7 +1429,7 @@ def get_agent_run_status(agent_run_id: str):
     run = frappe.db.get_value(
         "Agent Run",
         agent_run_id,
-        ["name", "agent", "status", "response", "error_message", "conversation"],
+        ["name", "agent", "status", "response", "error_message", "conversation", "owner"],
         as_dict=True,
     )
     if not run:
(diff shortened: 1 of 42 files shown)


esafwan:

PR #733, opened 2026-09-14
feat: allow alias/is_public/agent in HUF App manifests (guest portal Phase 1)

esafwan:
## Summary

Phase 1 of the HufAppDeskPortalDelivery track (closing item 1 of the "Net gap
list" in `doc/features/apps/delivery-portal-vs-desk.md`): the guest-facing
public app renderer (`/huf/apps/<alias>`, `huf.ai.app_public_renderer.py`)
already worked end-to-end, but a provider app had no way to declare
`alias`/`is_public`/`agent` through its manifest — a System Manager had to
hand-edit the `HUF App` Desk record after every sync, with no protection
against that edit being silently dropped on re-sync.

- Adds `alias`, `is_public`, `agent` to `ALLOWED_FIELDS` in
  `huf/ai/app_seeding/apps_loader.py`, with validation (`alias` must match
  the same slug shape as `app_id` and be unique; `agent` must name an
  existing `Agent` doc; `is_public` must be a JSON boolean).
- Extends the existing manual-override carve-out (`values.pop(...)` on
  update, same pattern `enabled` already has) to all three fields, so a
  System Manager's manual Desk edit survives re-sync — first-insert-only
  semantics.
- Extends the existing `app_id` collision-logging path to also validate
  `alias` uniqueness: a duplicate alias is rejected, logged under
  `"HUF App Registration Collision"`, and noted on the rejected app_id's
  own `sync_error`, without touching the record that legitimately holds the
  alias.
- Adds 7 unit tests in `huf/ai/app_seeding/tests/test_apps_sync.py` covering
  normalization, invalid `alias`/`is_public`/`agent`, first-insert
  semantics, manual-override survival on resync, and duplicate-alias
  rejection.

Only Phase 1 (backend manifest grammar) is in scope here. Phases 2-4
(per-app `www/` template registry, shared chat/agent core extraction, and
classic Desk `page.js` delivery) are real architecture decisions each
needing design sign-off first, and are intentionally not touched by this PR
— see `Tracks/safwan-erooth.HufAppDeskPortalDelivery/README.md`.

## Test plan

All verification was run live against a disposable bench
(`huf-desk-portal-p1`, branched off `pre-develop` at `f7062a11` — includes
PR #732) inside the project's devcontainer:

- `bench --site huf-desk-portal-p1.local run-tests --app huf --module huf.ai.app_seeding.tests.test_apps_sync`
  → `Ran 33 tests ... OK (skipped=23)` (the 23 skips are the pre-existing
  `RegressionCI` quarantine, unrelated to this change; all 7 new tests for
  alias/is_public/agent pass).
- Live manifest sync via `huf.ai.app_seeding.apps_loader.upsert_huf_app`
  in `bench console`, exercising the real validated code path end to end:
  - A manifest declaring `alias`, `is_public: true`, and `agent` synced
    successfully and persisted exactly as declared
    (`sync_status: Active`).
  - The exact DB lookup `app_public_renderer.py` uses to resolve
    `/huf/apps/<alias>` resolved the alias to the right `HUF App` + `Agent`,
    and `huf.ai.agent_access.check_agent_access(agent_doc, user="Guest")`
    returned `True` — confirming the guest portal path genuinely works
    end-to-end with these fields, not just that validation passes.
  - Manually flipped `is_public` to `0` on the record, then re-synced with
    a manifest that again asked for `is_public: true` plus a changed
    title: `is_public` stayed `0` (manual override survived) while the
    title update still applied.
  - A second manifest (different `app_id`) declaring the same `alias` was
    rejected with `Duplicate alias 'phase1verify': already registered by
    HUF App 'phase1verify' ...`, and the second `app_id` was never
    inserted at all.
- `bench --site huf-desk-portal-p1.local migrate` ran clean on the branch
  (no doctype/schema issues — `alias`/`is_public`/`agent` already exist on
  the `HUF App` DocType from prior work; this PR only changes the manifest
  grammar layer).

## Correction (2026-09-15)

The "guest portal path genuinely works end-to-end" claim above was **wrong** in one respect: the
verification above only exercised the DB lookup and `check_agent_access` inside `bench console` —
it never actually issued a real HTTP request to `/huf/apps/<alias>`. Doing that now (as part of the
Phase 2 track) surfaced two real bugs, both now fixed by commit `9cad6de7` on this branch:

1. **The frontend was never built on this bench**, so `huf/www/huf.html` (generated at build time
   by `frontend/scripts/copy-html-entry.mjs`) did not exist. `bench build --app huf` fixes this
   (environment gap, not a code bug — but it does mean the "end-to-end" claim above had not actually
   been checked against a running site).
2. **`app_public_renderer.py`'s `render()` called `TemplatePage("huf", ...).render()` unconditionally**,
   without first checking `can_render()`. When `huf.html` doesn't exist (as in the unbuilt state
   above), `TemplatePage.set_template_path()` never sets `self.app_path`, and calling `.render()`
   anyway crashes with `AttributeError: 'TemplatePage' object has no attribute 'app_path'` deep
   inside `set_pymodule()` — a real HTTP 500, not the graceful `PageDoesNotExistError` every other
   rejection in that method already returns. Fixed by guarding with `can_render()` first, matching
   `PathResolver`'s own flow.

Re-verified over real HTTP after both fixes:

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:8112/huf
200
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:8112/huf/apps/meeting-recorder
200
```

Both now return a real `200` with the SPA shell body, using the pre-existing `meeting-recorder`
HUF App (`alias=meeting-recorder`, `is_public=1`, `agent=Meeting Summary Agent` with
`allow_guest=1`) already present on this bench. The DB-level verification steps in the "Test plan"
above are still accurate as far as they went; they just weren't sufficient on their own to support
the "end-to-end" wording, which this correction retracts.


esafwan:
Consolidated into #737, which carries this phase plus the other three and the follow-up fix round as one branch. Closing in favor of #737.

PR #734, opened 2026-09-14
feat: per-app www/ template registry + route uniqueness (guest portal Phase 2)

esafwan:
## Summary

Phase 2 of the HufAppDeskPortalDelivery track (closing gap list items 2 and 6 of
`doc/features/apps/delivery-portal-vs-desk.md`): turns a manifest's `route` from inert launcher
metadata into something HUF actually serves, for whichever provider app declares a `www_template`.
Builds on Phase 1 (#733) — the `alias`/`is_public`/`agent` manifest fields are treated as already
present.

### Design decision: registry mechanism mirrors `find_seed_dirs()`

The core design question this phase had to answer first: how does a provider app declare "I ship a
portal page" without editing shared `huf/www/`? The answer follows the exact precedent
`huf.ai.app_seeding.scanner.find_seed_dirs()` already establishes for seed directories — a new
`www_template` manifest field names a bare directory (same slug shape as `app_id`/`alias`), and a
new `scanner.find_www_template_dir(source_app, www_template)` resolves it against the provider app's
own source tree, checked in the same two-location order:

1. `apps/<source_app>/<source_app>/www/<www_template>/` (package root — preferred)
2. `apps/<source_app>/www/<www_template>/` (app repo root)

This keeps the "HUF is never modified" contract `app-pattern.md` already documents for seed
directories: no shared-file edits, no per-app `website_route_rules` entry to remember, nothing to
register except the manifest field itself. A new `huf.ai.app_portal_renderer.HufAppPortalRenderer`
(a `page_renderer`, same family as the existing `AgentStreamRenderer`/`HufAppPublicRenderer`) matches
incoming requests against the `HUF App` registry directly — **the registry is the routing table** —
so declaring `www_template` in a manifest is the entire integration surface for a provider app.

I deliberately did **not** reuse Frappe's own `TemplatePage` for rendering, even though it already
searches every installed app's `www/` tree for a matching filename. `TemplatePage`'s search matches
by filename only, first match wins across *all* installed apps — there's no way to pin it to one
specific provider app, which matters once two apps could plausibly ship a same-named template
directory. `HufAppPortalRenderer` resolves the provider app directly via the registry (`source_app` →
`www_template` → `find_www_template_dir`) instead.

### Standardized boot context

Per the phase brief, this reuses (rather than reinvents) `www/huf.py`'s existing Guest-branching boot
pattern: I extracted `build_boot_context()` out of `get_context()` in `huf/www/huf.py` — same CSRF
[...]

PR #735, opened 2026-09-14
Phase 3: extract shared agent chat/transport core

esafwan:
Part of track TRK-20260914-7af5 (HufAppDeskPortalDelivery). Builds on Phase 1 (#733) and Phase 2 (#734), based off `pre-develop` at `f7062a11` (same base as both).

## What

Extracts the message send/receive + history-polling transport out of `huf/www/agent_chat.html` into a standalone module, `huf/public/js/agent-chat-core.js`, plain static JS served from `public/` (no bundle step needed — follows the same convention as the existing `huf-app-badge.js`/`huf-voice.js`), and retrofits `agent_chat.html` to consume it instead of its old inline copy.

## API shape

```js
const chat = HufAgentChatCore.mount({
  agent: "Support Agent",
  csrfToken: window.csrf_token,   // optional, falls back to the csrf_token cookie
  onMessage: (msg) => { ... },    // {role, kind, text|audioUrl, timeLabel}
  onLoading: (isLoading) => { ... },
  onError: (text) => { ... },
});

chat.sendText("hello");
chat.sendAudio(blob);
chat.refreshHistory();
chat.reset();
```

No page-load-time globals or singleton state — `mount()` returns an instance-scoped object, so more than one instance can exist on a page, and a caller can create/destroy one per page-show. This is deliberate for the two known future consumers:

- **Phase 2's portal template** (`huf.ai.app_portal_renderer`, already merged in #734): its Jinja template can `<script src="/assets/huf/js/agent-chat-core.js">` and call `mount()` with the boot context it already assembles via `www/huf.py`'s `build_boot_context()` pattern. No changes needed to #734 for this to work — the module's `endpoints` option exists specifically so a consumer isn't hardcoded to `huf.ai.agent_chat.*` if a future surface needs different whitelisted methods.
- **Phase 4's Desk page.js** (next): `frappe.require('/assets/huf/js/agent-chat-core.js')` then `mount()` inside a `frappe.ui.Page`'s body, with `destroy()` called (no-op today, reserved) on page hide.

## Why not an ES module / bundled asset

`agent_chat.html` and a Desk `page.js` are both loaded outside the SPA's Vite bundle (one is a static Jinja/HTML page with no build step, the other goes through `frappe.require` per `assets/asset-bundling.md`'s "Lazy loading in /app"). A plain global (`window.HufAgentChatCore`) via a `<script src>` tag is what both can consume without adding a bundler dependency; the module also assigns `module.exports` under a CommonJS/ESM wrapper for any future consumer that does go through a bundler.

## Test plan / live verification

Verified on bench `huf-desk-portal-p1` (`localhost:8112`):

- `bench build --app huf` succeeds; `/agent_chat.html` and `/assets/huf/js/agent-chat-core.js` both serve 200, and the page references the new script tag.
- Loaded `/agent_chat.html` in a real browser (Chrome via the Browser pane) as Guest: page renders with no console errors, typed and submitted a text message, confirmed the module's `onMessage` hook immediately rendered the user bubble and made the real `POST /api/method/huf.ai.agent_chat.create_conversation` call — this 403'd as Guest, which is expected and matches the pre-refactor page's behavior identically (`create_conversation`/`send_message_to_conversation` are not `allow_guest`; this was never callable by a Guest before this change either).
- To verify the authenticated path without entering credentials into any UI, logged in via `POST /api/method/login` with a bench-local test admin password and replayed the exact three calls the module makes (`create_conversation`, `send_message_to_conversation`, and the `Agent Message` history query) directly against the bench over HTTP:
  - `create_conversation` → `{"success":true,"conversation_id":"agok3l6vrt"}`
  - `send_message_to_conversation` → the user's message was persisted as an `Agent Message` doc (confirmed via the history query, `content` matches exactly what was sent). The agent's own reply generation failed on this bench with `Password not found for AI Provider Google api_key` — a pre-existing bench configuration gap (no LLM provider key set on this disposable bench), unrelated to this change; the transport request/response cycle itself completed correctly.
- No regressions to `agent_chat.html`'s presentation: message bubbles, audio player, mic/waveform recording UI are all unchanged and still owned by the page's own script, which now only consumes the core for network calls.

## Lint / tests

No root-level JS test harness applies to this file (it's a plain unbundled `public/js/` asset, not part of `frontend/`'s Vitest/Playwright suite). `frontend/`'s own `package.json` has `test`/`test:e2e` scripts but they only cover the Vite-built SPA, not `www/`-served static pages. The repo's `.eslintrc` (root, `root: true`) is written for the legacy ESLint config format; the only installed ESLint binary in this checkout is `frontend/node_modules/.bin/eslint` (v10, flat-config only) and is incompatible with the root `.eslintrc`, so it can't lint this file with the tooling actually present on this bench — flagging this rather than fabricating a lint pass. Verified the new module's syntax directly with `node -c huf/public/js/agent-chat-core.js`.

No Python was touched in this PR.


esafwan:
Consolidated into #737, which carries this phase plus the other three and the follow-up fix round as one branch. Closing in favor of #737.

